Issue #2, opened 2018-06-27
Files with extremely long filenames do not get packaged


abenson:
It looks like files with paths over 127 characters will not get packaged. 

I created two directory structures, 1 with 127 characters, another with 128. The 127 character path gets added, the 128 character does not.

~~~
greenleaf% a=$(echo a{1..33} | sed 's| |/|g')
greenleaf% mkdir -p $a
greenleaf% touch $a/aaa
greenleaf% echo $a/aaa | wc -c
127
greenleaf% b=$(echo b{1..33} | sed 's| |/|g')
greenleaf% mkdir -p $b
greenleaf% touch $b/bbbb
greenleaf% echo $b/bbbb | wc -c
128
greenleaf% find . -type f
./a1/a2/a3/a4/a5/a6/a7/a8/a9/a10/a11/a12/a13/a14/a15/a16/a17/a18/a19/a20/a21/a22/a23/a24/a25/a26/a27/a28/a29/a30/a31/a32/a33/aaa
./b1/b2/b3/b4/b5/b6/b7/b8/b9/b10/b11/b12/b13/b14/b15/b16/b17/b18/b19/b20/b21/b22/b23/b24/b25/b26/b27/b28/b29/b30/b31/b32/b33/bbbb
greenleaf% cd ..
greenleaf% xbps-create -A noarch -n test-1.0_1 -s "testing filename length" testdest 
test-1.0_1: adding `/a1/a2/a3/a4/a5/a6/a7/a8/a9/a10/a11/a12/a13/a14/a15/a16/a17/a18/a19/a20/a21/a22/a23/a24/a25/a26/a27/a28/a29/a30/a31/a32/a33/a' ...
test-1.0_1: binary package created successfully (test-1.0_1.noarch.xbps)
greenleaf% tar Jtvf test-1.0_1.noarch.xbps
-rw-r--r-- root/root       496 1969-12-31 18:00 ./props.plist
-rw-r--r-- root/root     11652 1969-12-31 18:00 ./files.plist
-rw-r--r-- 0/0               0 2018-06-27 07:59 ./a1/a2/a3/a4/a5/a6/a7/a8/a9/a10/a11/a12/a13/a14/a15/a16/a17/a18/a19/a20/a21/a22/a23/a24/a25/a26/a27/a28/a29/a30/a31/a32/a33/a
~~~

alexjarosch:
[...]

PR #4, opened 2018-06-28
simple warning for the 128 char limit in filenames incl. paths

alexjarosch:
I think it is crucial to  be aware of the problem with 128 char limits in the current implementation of xbps-create, as described in issue https://github.com/void-linux/xbps/issues/2.

Extending the portableproplib to accept keys with more than 128 characters is not trivial as I found out today, neither is fixing the current xbps-create implementation.

Until we have fixed this I think it is important to keep an eye on what is not getting packed into a package with xbps. This could have quite some stability implications for void-linux and currently nobody knows how much is not getting packed when using xbps-create.

Utilizing the test from @abenson in issue https://github.com/void-linux/xbps/issues/2 results now in a warning message:
`! Warning not included (128char bug): /b1/b2/b3/b4/b5/b6/b7/b8/b9/b10/b11/b12/b13/b14/b15/b16/b17/b18/b19/b20/b21/b22/b23/b24/b25/b26/b27/b28/b29/b30/b31/b32/b33/bbbb`
and as he demonstrated, the bbbb file does not get packed into the xbps package with the current version of xbps-create.

Hope this helps until we fix the issue

### DIFF
--- a/bin/xbps-create/main.c
+++ b/bin/xbps-create/main.c
@@ -319,6 +319,11 @@ ftw_cb(const char *fpath, const struct stat *sb, int type, struct FTW *ftwbuf _u
 	 * use the keysym, as this value has the unsanatized path.
 	 */
 	xbps_dictionary_set_cstring(fileinfo, "file", filep);
+
+	/* Warning due to 128 char limit in file paths */
+	if (strlen(filep) >= 128)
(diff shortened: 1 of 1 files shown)


alexjarosch:
This should be fixed now (https://github.com/void-linux/xbps/issues/2), no need to pull the simple fix above.

PR #15, opened 2018-08-23
Fix namespace violation

foxlet:
Fixes building on PowerPC systems (depends on header files that conflict with _unused).

Based on patch by @ncopa from the previous repository: voidlinux/xbps#292

### DIFF
--- a/bin/xbps-alternatives/main.c
+++ b/bin/xbps-alternatives/main.c
@@ -55,7 +55,7 @@ usage(bool fail)
 }
 
 static int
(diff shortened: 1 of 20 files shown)


Vaelatern:
lgtm, @Duncaen can you think of any downsides?

ncopa:
Note that identifiers that starts with underscore is reserved, so this change is correct. If you are unsure that `UNUSED` is conflicting with anything, then you can use `XBPS_UNUSED` or similar.

PR #22, opened 2018-10-05
lib/package_unpack.c: remove files of wrong filetype in path

### LINKED ISSUE #14: xbps-install fails to replace file with directory on update
`xbps-install` fails to replace a file with a directory of the same name when updating packages.
Here is an example to trigger the bug: https://github.com/jnbr/void-packages/commit/545a737bd7ff2f1110bdd83712fe8e9f5681fcfe

This is similar to: https://github.com/voidlinux/xbps/issues/220

As a workaround, the affected package can be removed and installed afterwards.

jnbr:
This change removes regular files which prevent extracting directories with the same name. Fixes #14.

I only tested it with updating gimp to version 2.10.6_1.

It doesn't fix the other way round where a regular files has to be extracted over a non-empty directory (this fails with `[unpack] failed to extract files: Directory not empty`).

### DIFF
--- a/lib/package_unpack.c
+++ b/lib/package_unpack.c
@@ -75,6 +75,7 @@ unpack_archive(struct xbps_handle *xhp,
 {
 	xbps_dictionary_t binpkg_propsd, binpkg_filesd, pkg_filesd;
 	xbps_array_t array, obsoletes;
+	xbps_object_iterator_t iter;
 	xbps_object_t obj;
 	xbps_data_t data;
 	const struct stat *entry_statp;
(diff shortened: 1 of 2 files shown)


CameronNemo:
Hi, have you tested how this interacts with conffiles?

jnbr:
> Hi, have you tested how this interacts with conffiles?

I tested now. I doesn't remove conffiles, so updating a a pckage where a configfile becomes a directory will still brake, which is the better option here imo.

While testing I found a bug when using `-r` which is fixed now.

luzpaz:
ETA on this?  
Asking because GIMP 2.10.8 update relies on this

CameronNemo:
@luzpaz have you tested this? Unfortunately I haven't had a chance to.

CameronNemo:
I was running this for at least a couple weeks, didn't notice anything wrong.

jnbr:
I discussed this PR with @Duncaen.
There are similar issues with symlinks: https://github.com/voidlinux/xbps/pull/221
We agreed that it would be better to remove obsolete files before extracting new files, so we don't need special handling for many similar but different cases.

PR #28, opened 2018-11-05
Fixed typo in die string: uidmap -> uid_map.

parke:


Duncaen:
You can already build xbps on non-void systems?

parke:
Oh, sorry.  I did not realize I had submitted that commit as a pull request.  I am just beginning to learn to use Github to contribute to projects.  I did not realize that each additional push to the same branch would be part of the same pull request.  Had I realized that, I would have created another branch.

(My non-Void system is Ubuntu, so henceforth I will say Ubuntu instead of non-Void.)

I have not seriously tried to build xbps on Ubuntu.  I imagine it would be tricky.

However, I only need to build xbps-uunshare.  Then I can use xbps-uunshare to chroot into a Void ROOTFS (extracted from the official Void ROOTFS tarball).

It is fairly easy to build xbps-uunshare on Ubuntu, as you can see from the above commit.

There is one additional problem.

I first run xbps-uunshare to enter the ROOTFS.  I then run xbps-src.  xbps-src runs xbps-uunshare a second time (this time inside the chroot).  This second invocation of xbps-uunshare fails.  The call to unshare(CLONE_NEWUSER) fails inside a chroot.  See the EPERM section of the unshare(2) manpage for details.

I am currently working on another commit that will allow xbps-uunshare to work inside an xbps-uunshare chroot.  I am not sure if I will finish that commit today, but I do not expect it to be a big commit.

My goal is to be able, as a non-root user on Ubuntu, to chroot into Void and then build Void packages using xbps-src.

Vaelatern:
If you would like, you can use `git rebase -i` to remove commits from a branch (remember to save those commits in a new branch first!) and `git push -f` to fix the pull request.

xtraeme:
This can be closed, was fixed by @Duncaen as shown above.

Issue #23, opened 2018-10-14
Empty directories don't get removed on xbps-remove


steinex:
When deinstalling a package, empty directories that were created by the package on installation are left behind.

For example, `cmus` leaves `/usr/lib/cmus`.
Another: `faenza-icon-theme` leaves `/usr/share/icons/Faenza*`.

It's not so that every directory is left behind, though. In the cmus example, `/usr/share/cmus` is properly cleaned up. I can't see a pattern in which case a directory is deleted and in which not.

It seems not to be package specific.

Vaelatern:
Seems interesting, `/usr/lib/cmus` is itself only home to two separate directories, `op` and `ip`.
@steinex please clarify: which dirs in `faenza-icon-theme` are left behind?

steinex:
```
/usr/share/Faenza
/usr/share/Faenza-Ambiance
/usr/share/Faenza-Dark
/usr/share/Faenza-Darker
/usr/share/Faenza-Darkest
/usr/share/Faenza-Radiance
```

kitty is another example, it leaves:
```
/usr/lib/kitty
/usr/lib/kitty/kittens
/usr/lib/kitty/kitty
/usr/lib/kitty/terminfo
```

biopsin:
/usr/lib/ImageMagick-7.0.8 - a subfolder "modules-Q16HDRI" blocks the removal

Vaelatern:
But it's not just subfolders. As best I can tell, `/usr/share/Faenza-Radiance` has no subfolders, just files.

jnbr:
There is no `/usr/share/Faenza-Radiance` at all in `faenza-icon-themes`. It's all in `/usr/share/icons/Faenza*`

steinex:
Yes, this was my error, sorry. The above listing of mine for `faenza-icon-themes` misses `/icons/`. However, it was just an example anyway as the issue applies to all packages and still stands.

jnbr:
this are at least two different bugs:

for `kitty` and most recently built packages, the problem is that the order of entries in `files.plist` was changed in 3ef932de2f86b239bc7b0d5d3b8784b3c2ebdb74. This change requires to remove the directories listed in `files.plist` in reverse order.

`faenza-icon-theme` was built `2015-08-05 13:56 UTC`, way before the change. `xbps-remove --debug faenza-icon-theme` complains about modified symlinks and leafs them in place.

xtraeme:
This should be fixed in master, see #31

[...]

PR #30, opened 2018-11-13
lib/package_remove: reverse order of file removal

jnbr:
`xbps-remove` fails to remove some directories, see https://github.com/void-linux/xbps/issues/23.
This happens because they are removed in the wrong order, e.g. for a package with the files `a` and `a/b`, `remove(3)` is first called on `a` where it fails and then on `a/b`.

Reversing the order of the files to be removed fixes this.

### DIFF
--- a/lib/package_remove.c
+++ b/lib/package_remove.c
@@ -100,7 +100,6 @@ remove_pkg_files(struct xbps_handle *xhp,
 		 const char *pkgver)
 {
 	xbps_array_t array;
-	xbps_object_iterator_t iter;
 	xbps_object_t obj;
 	const char *curobj = NULL;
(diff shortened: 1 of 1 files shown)


jnbr:
For the examples I looked at, it was the case and it seems to be the case for all packages created with the current version of `xbps-create`, I'm not sure about earlier versions of `xbps`.

Vaelatern:
Is it sorted by accident, then?

jnbr:
The sorting is caused by `walk_dir()` in `bin/xbps-create/main.c` which does a postorder traversal.
Earlier version used `nftw(3)`, which does a preorder traversal.

So there is no guaranteed order. A more robust implementation would be to order the file paths by length before removal.

PR #31, opened 2018-11-16
lib/package_remove: order directories before removal

jnbr:
The order of entries in files.plist is inconsistent between different releases of xbps.
Hence they must be ordered before removal to, so sub directories are removed before their parents.

This solves a part of #23.

### DIFF
--- a/lib/package_remove.c
+++ b/lib/package_remove.c
@@ -93,13 +93,23 @@ check_remove_pkg_files(struct xbps_handle *xhp,
 	return fail;
 }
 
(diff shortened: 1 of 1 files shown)


joshuakraemer:
@jnbr, thank you for your patch. Is there a reason why this can't be merged?

jnbr:
> Is there a reason why this can't be merged?

This needs at least some review before it can be merged. There is probably a better way to fix this.

xtraeme:
Please repush

PR #36, opened 2018-12-13
travis: downgrade libarchive and fix travis config

bobertlo:
I wanted to fix travis. Ubuntu only has libarchive up to 3.1.2. I'm not sure if there is anything that strongly depends on >=3.2.0, but all the tests pass now. I suppose the only other option would be to compile a newer version on travis. There was a long gap between 3.1.2 and 3.2.x, and looks like they found some security bugs, but in practice xbps will be built on void and with 3.3.x anyway. I can build a newer version from source if that is desired, though.

I also had to slightly change some of the travis scripts to get all the dependencies to compile and work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ compiler:
 
 addons:
   apt:
-    sources:
+    packages:
         - gdb
         - liblua5.2-0
         - liblua5.2-dev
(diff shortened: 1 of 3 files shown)


bobertlo:
Closing this, working on a better solution.

PR #37, opened 2018-12-13
xbps-install: add option to only download packages

CameronNemo:
Rebase of #9


Vaelatern:
Instead I rebased #9 directly.

PR #38, opened 2018-12-13
import gcc8 patches from xbps-src

bobertlo:
These patches allow xbps to compile under gcc8. just pulled from void-packages.

### DIFF
--- a/tests/xbps/libxbps/config/main.c
+++ b/tests/xbps/libxbps/config/main.c
@@ -39,6 +39,8 @@ ATF_TC_BODY(config_include_test, tc)
 	struct xbps_handle xh;
 	const char *tcsdir;
 	char *buf, *buf2, pwd[PATH_MAX];
+	char confdir[PATH_MAX + 1 + sizeof("xbps.d")];
+	size_t len;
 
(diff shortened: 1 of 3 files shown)


CameronNemo:
the xbps-checkvers one is -1

the other one with relpath is weird

the other too

snprintf and then memcpy seems useless

just check the return value of snprintf and error out if its truncated

tests/xbps/libxbps/config/main.c
```diff
@@ -39,6 +39,8 @@ ATF_TC_BODY(config_include_test, tc)
 	struct xbps_handle xh;
 	const char *tcsdir;
 	char *buf, *buf2, pwd[PATH_MAX];
+	char confdir[PATH_MAX + 1 + sizeof("xbps.d")];
```
CameronNemo: why allow confdir to be larger than PATH_MAX? why not cap it?

Vaelatern: I suspect for the null terminator at least

bobertlo:
Closing this and deferring to #58 and #59

PR #39, opened 2018-12-14
travis: run void host in docker

bobertlo:
The current CI runs in ubuntu containers, which do not provide a recent enough version of libarchive. This uses official Void docker images to compile and run tests on the xbps repo.


bobertlo:
I rebased this to remove the xbps-src patches, anticipating better patches to actually be merged. (#58 #59)

the-maldridge:
This could do with some more documentation in terms of what the problem is and how this resolves it.

bobertlo:
@the-maldridge I updated the description. This was in response to #36 trying to make it work on ubuntu hosts.

bobertlo:
I just made it actually run clang (which does not error) and removed `travis-install-deps.sh` now.

bobertlo:
@Duncaen @Vaelatern any thoughts on this?

the-maldridge:
Looks okay to me, but I'd like to have someone who deals with xbps more take a look.  @Duncaen what do you think of running the xbps tests from the void environment?

bobertlo:
I tested this again with the newly integrated docker images. Still waiting on the gcc8 fixes, but working otherwise. Any thoughts @Duncaen ?

Duncaen:
Should be fine as long as its not using the the same configure paths as we use in void (breaks the test suite if configured with `--prefix=/usr` on a system with configurations files in `/usr/share/xbps.d`).

Duncaen:
Hm maybe add some more make jobs.

bobertlo:
@Duncaen Ok, set it to use 4 makejobs (based off xbps-src, I don't know what it's running on.)

CLang is running all the test suite, GCC when I tested with the hacky gcc8 patches.

PR #44, opened 2019-01-09
Don't process processed packages again in collect_rdeps

yshui:
Otherwise we will get exponential time complexity


Chocimier:
Looks correct. How it was tested?

yshui:
@Chocimier I just ran the test suite. 

Looks like the CI is not set up properly?

PR #46, opened 2019-01-11
initend.c: properly free result of scandir()

huglovefan:
it frees each entry as it loops through them, but the `break` in the loop could leave the rest unfreed

i've built xbps with this patch using xbps-src and valgrind now reports no leaks for `xbps-query pkg` (and one less leak for programs using libxbps)

### DIFF
--- a/lib/initend.c
+++ b/lib/initend.c
@@ -316,32 +316,34 @@ parse_dir(struct xbps_handle *xhp, const char *cwd, const char *confdir, const c
 	 */
 	xbps_dbg_printf(xhp, "Processing configuration directory: %s\n", confdir);
 
+	namelist = NULL;
 	if ((n = scandir(confdir, &namelist, 0, alphasort)) < 0)
 		goto stage2;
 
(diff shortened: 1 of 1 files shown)


Vaelatern:
Please add a sanity check to make sure `namelist[i]` is non-NULL.

huglovefan:
like `if (namelist[i] != NULL) free(namelist[i]);` or something else?

is it not enough to assume scandir() isn't broken and `free(NULL)` is a no-op?

it's already checking the return value of scandir and accessing `namelist[i]->d_name` for some of the entries

xtraeme:
Please repush

xtraeme:
Hold on, because I'm not sure this will be effective with the conf refactoring code by duncaen

xtraeme:
ping?

Vaelatern:
Your comment was the latest.

huglovefan:
not needed anymore with https://github.com/void-linux/xbps/pull/63

PR #47, opened 2019-01-16
fix setting alternatives groups for packages that have more than one

huglovefan:
the `group && strcmp(keyname, group)` check in the loop would make the function always return `ENOENT` when setting an alternatives group that wasn't the only/last one in the dictionary

`xbps-alternatives` checks the return value of `xbps_alternatives_set` and doesn't update the pkgdb on failure (even if it updated the symlinks)

you can test it with `xbps-alternatives -g hostname -s busybox` and `xbps-alternatives -g hostname -l` (assuming you have hostname from another package)


huglovefan:
i've updated an existing test to fail without the change if that's okay

xtraeme:
I'd like to see this one in 0.54.

PR #49, opened 2019-01-23
xbps-create: handle unsupported files

Duncaen:


Duncaen:
Merged as https://github.com/void-linux/xbps/commit/a46be6867a41614389fb3edd6039446c21a9a6f6.

PR #50, opened 2019-01-24
Maybe we need upgrade to openssl 1.1.x.

wuhanck:
Maybe people are right. Openssl do make their bad API worse.
But still we need to live with it.
Lot of distro have been upgraded to openssl 1.1.x.
It is better to support 1.1.x and drop 1.0.x support. (Or add some FENCE to make both work?)
May somebody help and pick the modification and make xbps works for 1.1.x?

### DIFF
--- a/configure
+++ b/configure
@@ -678,7 +678,7 @@ fi
 # libssl with pkg-config support is required.
 #
 printf "Checking for libssl via pkg-config ... "
-if $PKGCONFIG_BIN --exists 'libssl < 1.1' && ! $PKGCONFIG_BIN --exists libtls ; then
+if $PKGCONFIG_BIN --exists 'libssl < 1.2' && ! $PKGCONFIG_BIN --exists libtls ; then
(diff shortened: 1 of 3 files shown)


Vaelatern:
@void-linux/pkg-committers Do we want to support xbps bootstrapping without libressl? If so, we probably need this....

maxice8:
On Thu, Jan 24, 2019 at 11:30:54AM -0800, Toyam Cox wrote:
> @void-linux/pkg-committers Do we want to support xbps bootstrapping without libressl? If so, we probably need this....
> 
> -- 
> You are receiving this because you are on a team that was mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/void-linux/xbps/pull/50#issuecomment-457326253

I wouldn't invest time into supporting alternatives to LibreSSL but if people want to do it as long as it doesn't cause problems for LibreSSL i don't see a problem.

PR #54, opened 2019-02-07
lib/fetch/common.c: send socks5 request as a single transaction

andkem:
Splitting the socks5 request is causing issues when running against Tor.
It causes Tor to fail with an address unknown error. Assembling the
request and sending it in its entirety solves the issue.

An strace of torsocks, that function correctly when used with xbps shows
that the request is being sent all at once:

        recvfrom(3, "\5\0", 2, 0, NULL, NULL)
        sendto(3, "\5\1\0\3>lysator7eknrfl47rlyxvgeamrv7ucefgrrlhk7rouv3sna25asetwid.onion\0P", 69, 0, NULL, 0)
        recvfrom(3, "\5\0\0\1\0\0\0\0\0\0", 10, 0, NULL, NULL)

### DIFF
--- a/lib/fetch/common.c
+++ b/lib/fetch/common.c
@@ -274,11 +274,19 @@ fetch_bind(int sd, int af, const char *addr)
 int
 fetch_socks5(conn_t *conn, struct url *url, struct url *socks, int verbose)
 {
-	char buf[16];
+	char buf[262];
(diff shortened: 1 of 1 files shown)


xtraeme:
Please repush

xtraeme:
@Duncaen please review

PR #58, opened 2019-02-23
lib/initend.c: xbps_init(): safer string handling

CameronNemo:
This adds return value checks for strdup() and snprintf() invocations in xbps_init().

Related to #38 

Signed-off-by: Cameron Nemo <camerontnorman@gmail.com>


Duncaen:
I would remove the `size` variable and use `sizeof` direct.

CameronNemo:
Reasons I don't put the sizeof in the snprintf call:

1. The line wrapping is already bad, I would rather not make it worse.
2. The sizeof and snprintf match up perfectly, so the sizeof argument is the same as the destination.
3. Less repetition means less places to modify in the future.

PR #60, opened 2019-02-23
README: update voidlinux URLs

bobertlo:


Vaelatern:
I think this is OK, but want an annotation clarifying this is just one mirror.

bobertlo:
@Vaelatern does this work?

PR #62, opened 2019-03-05
Add the ability to ignore packages

Duncaen:
The previous idea was to use virtual packages in the users configuration
to satisfy dependencies by mapping them to existing installed packages.
Using virtual packages for it doesn't work as expected and trying to make
it work would break other functionalities of virtual packages, like the
version satisfaction checks for `provides` and the ability to replace
virtual packages with real packages. The virtual package functionality
should be used exclusively for virtual packages.

This allows users to specify packages packages that should be ignored.
Ignored packages in dependencies are always satisfied without installing
the package, while updating or installing a package that depends on an
ignored package.

This does NOT ignore the shlib checks, ignoring a package that provides
required shared libraries will abort the transaction as if there was no
package that provides the required shared library.

From `xbpd.d(5)`:
```
     ignorepkg=pkgname
         Declares a ignored package.  If a package depends on an ignored
         package the dependency is always satisfied, without installing the
         ignored package.
```


Vaelatern:
This looks good!

xtraeme:
Please repush!

xtraeme:
This one can be closed now, merged into master already.

PR #63, opened 2019-03-06
Refactor configuration parsing

Duncaen:


lib/log.c
```diff
@@ -0,0 +1,102 @@
...
+	if ((xhp->flags & XBPS_FLAG_DEBUG) == 0)
+		return;
+
+	va_start(ap, fmt);
+	common_printf(stderr, "[DEBUG] ", fmt, ap);
+	va_end(ap);
+}
+
+void
+xbps_error_printf(const char *fmt, ...)
```
CameronNemo: What does a function offer here?

```
#define xbps_error_printf common_printf(stderr, "ERROR: ", __VA_ARGS__)
#define xbps_warn_printf common_printf(stderr, "WARNING: ", __VA_ARGS__)
```

lib/conf.c
```diff
@@ -0,0 +1,383 @@
+/*
+ * Copyright (c) 2011-2015 Juan Romero Pardines.
```
CameronNemo: Why not just convert to SPDX for this file and future files?

Vaelatern: You need to copy these sorts of licenses with the source.

lib/log.c
Duncaen: They are exported symbols in libxbps, I don't think this wouldn't make a big difference and would break ABI.

lib/conf.c
CameronNemo: I would still suggest adding SPDX to make processing easier.

lib/conf.c
```diff
@@ -0,0 +1,383 @@
...
+ * Copyright (c) 2014 Enno Boland.
+ * Copyright (c) 2019 Duncan Ovebrbruck <mail@duncano.de>.
```
xtraeme: Typo? 😅

Duncaen:
I think this is not ready yet, there are some test cases missing i.e. for absolute includes which should be broken with this commit.

xtraeme:
We need test cases for everything

xtraeme:
That's why I suggested to repush. I think the current set of tests shall find any issue. But perhaps we are missing new ones. Please verify, thanks.

xtraeme:
ping

Issue #65, opened 2019-03-08
Show action hold for pkg downgrade 


biopsin:
Holding a package that has action downgrade, the action description does not change, 
however download size is empty.
Holding a package should show action hold for that package.

PR #67, opened 2019-03-15
 bin/xbps-install: don't overwrite action with downgrade if its on hold 

Duncaen:
For #65


bin/xbps-install/util.c
```diff
@@ -158,7 +158,7 @@ print_trans_colmode(struct transaction *trans, int cols)
 		ver = xbps_pkg_version(pkgver);
 		if (iver) {
 			int rv = xbps_cmpver(iver, ver);
-			if (rv == 1)
+			if (rv == 1 && strcmp(tract, "hold") != 0)
```
xtraeme: strcmp(tract, "hold"))

Duncaen:
Merged with suggested change.

PR #68, opened 2019-03-16
Store file, directory sizes in files.plist and add directory_sizes dictionary to repodata

Duncaen:
The first commit makes `xbps-create` store the size of files and directories (sum of all files and sub directories inside the directory) in the `files.plist` by adding a new `size` key.

The second commit uses the newly added data to create a dictionary of directories stored in the repository data. 

The new dictionary in the repodata looks something like this:

```
<key>directory_sizes</key>
<dict>
    <key>/bar</key>
    <integer>27</integer>
</dict>
```

The root directory is not in the dictionaries, the size is already available as `installed_size`.

The purpose of this change is to later add per mount point available size checks.

`repodata` needs to contain the directory sizes because xbps asks for confirmation of the transaction and does the filesystem size checks before downloading the packages.

### DIFF
--- a/bin/xbps-create/main.c
+++ b/bin/xbps-create/main.c
@@ -61,6 +61,7 @@ struct xentry {
 	uint64_t mtime;
 	char *file, *type, *target, *hash;
 	ino_t inode;
+	size_t size;
 };
 
 static TAILQ_HEAD(xentry_head, xentry) xentry_list =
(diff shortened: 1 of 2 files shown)


CameronNemo:
>repodata needs to contain the directory sizes because xbps asks for confirmation of the transaction and does the filesystem size checks before downloading the packages.

Why not do two size checks -- one for /var/cache/xbps and one after downloading for unpacked size?

Duncaen:
Good idea, I think that would work too.

I can move the second space check to `xbps_transaction_commit`.
The first check would instead of checking the whole size on root only check for `dlsize` space in `cachedir`.
Somewhere in `xbps_transaction_commit` the second check is done and can then abort/return `ENOSPC`.

I think I keep the `xbps-create` change, the size value could be a useful addition for some other fast paths.
And it also reduces the number of paths that need to be mapped to mountpoints at update/install time and doesn't slow down `xbps-create` because it gets the stats for each file anyways.

PR #69, opened 2019-03-18
lib/fetch: add happy eyeballs connect algorithm (RFC8305)

Duncaen:
Connect to the addresses from `getaddrinfo(3)`,
alternating between address family,
starting with ipv6 and wait `fetchConnDelay`
between each connection attempt.

If a connection is established within the attempts,
use this connection and close all others.

If `connect(3)` returns `ENETUNREACH`, don't attempt more
connections with the failing address family.

If there are no more addresses to attempt,
wait for `fetchConnTimeout` and return the first established
connection.

If no connection was established within the timeouts,
close all sockets and return -1 and set errno to
`ETIMEDOUT`.


Duncaen:
I think I'm going to review this again and try to clarify the different states and maybe move some things into smaller functions.

xtraeme:
I think it's good as is (assuming it does what's expected)

xtraeme:
merged in master, please close.

PR #70, opened 2019-03-19
unpack: keep conf_files replaced with symlinks, unpack as .new-pkgver

Duncaen:
Before this patch configuration files that where not regular files, i.e. symlinks would always be removed.
This adds a test case and makes `xbps-install` always install configuration files as `.new-pkgver` if the configuration file was replaced with a symlink on disk. 

https://github.com/voidlinux/xbps/issues/276

### DIFF
--- a/lib/package_config_files.c
+++ b/lib/package_config_files.c
@@ -63,7 +63,8 @@ xbps_entry_install_conf_file(struct xbps_handle *xhp,
 			     xbps_dictionary_t pkg_filesd,
 			     struct archive_entry *entry,
 			     const char *entry_pname,
-			     const char *pkgver)
+			     const char *pkgver,
+			     bool symlink)
 {
 	xbps_object_t obj, obj2;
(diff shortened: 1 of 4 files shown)


lib/package_config_files.c
```diff
@@ -63,7 +63,8 @@ xbps_entry_install_conf_file(struct xbps_handle *xhp,
 			     xbps_dictionary_t pkg_filesd,
 			     struct archive_entry *entry,
 			     const char *entry_pname,
-			     const char *pkgver)
+			     const char *pkgver,
+				 bool symlink)
```
xtraeme: Please fix indent

xtraeme:
I'd like to see this one in 0.54.

Duncaen:
Merged https://github.com/void-linux/xbps/commit/ea2cb1d369b2d4afcde61ace59c594a06355ded2.

PR #74, opened 2019-04-02
data: update repo urls to voidlinux.org

jnbr:


the-maldridge:
Please use a-hel-fi instead.

jnbr:
done

PR #81, opened 2019-04-20
fulldeptree: make sure there are no dups.

xtraeme:
Added a test case to verify that dup deps are discarded
completely.


xtraeme:
This will success when https://github.com/void-linux/void-packages/pull/11236 is applied, and new pkgs are available.

PR #87, opened 2019-04-24
configure: added `--enable-rpath` option.

xtraeme:
If set, binaries will have rpath set to $ORIGIN/../lib. That means
that even using defaults (--prefix=/usr), you can install
it via DESTDIR anywhere in your system and binaries will work.

This is in preparation for BSD support.


xtraeme:
Ping

PR #89, opened 2019-05-02
xbps-install: make question() read the whole input line

huglovefan:
this fixes a bug:

1. ```sh
   mkdir test
   xbps-install -S -R https://alpha.de.repo.voidlinux.org/current -r test base-system
   ```
2. when it asks if you want to import the rsa key, type `y` and press enter
3. notice how it assumes yes when it asks if you want to install the packages

step 2 only reads the "y" character from stdin, so in step 3 it gets the newline and assumes the default answer

### DIFF
--- a/bin/xbps-install/question.c
+++ b/bin/xbps-install/question.c
@@ -41,22 +41,27 @@ static bool
 question(bool preset, const char *fmt, va_list ap)
 {
 	int response;
+	bool rv = false;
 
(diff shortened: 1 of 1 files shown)


xtraeme:
could you please repush to update travis result?

xtraeme:
ping?

xtraeme:
This one is also ready, and it's in the queue for 24 days already.

PR #91, opened 2019-05-14
Update revdeps when installing or updating pkgs

xtraeme:
This fixes the cases where a pkg installation or update
could break its revdeps due to not satisfying dependencies.

When performing an install or update, xbps now always checks if
there are updates for its revdeps and get added to the transaction.

See the two new test cases for more information.


CameronNemo:
Bumping copyright makes it harder to backport the commits. Usually better if those changes are in their own commit.

PR #92, opened 2019-05-15
package_remove: add test and fix length comparison

jnbr:
follow up of #31 
- added test case
- fixed comparison function


xtraeme:
Great! thanks :-)

PR #93, opened 2019-05-17
libxbps: remove dangling symlinks properly.

### LINKED ISSUE #23: Empty directories don't get removed on xbps-remove

xtraeme:
This fixes removal of packages that contain multiple levels
of dangling symlinks, i.e faenza-icon-theme and probably others.

Close #23

### DIFF
--- a/lib/package_remove.c
+++ b/lib/package_remove.c
@@ -190,7 +190,10 @@ remove_pkg_files(struct xbps_handle *xhp,
 		bool found;
 
 		xbps_dictionary_get_cstring_nocopy(obj, "file", &file);
-		snprintf(path, sizeof(path), "%s/%s", xhp->rootdir, file);
+		if (strcmp(xhp->rootdir, "/") == 0)
+			snprintf(path, sizeof(path), "%s", file);
+		else
+			snprintf(path, sizeof(path), "%s%s", xhp->rootdir, file);
 
 		if ((strcmp(key, "files") == 0) ||
 		    (strcmp(key, "conf_files") == 0)) {
--- a/lib/util.c
+++ b/lib/util.c
@@ -498,7 +498,7 @@ char *
 xbps_symlink_target(struct xbps_handle *xhp, const char *path, const char *tgt)
 {
 	struct stat sb;
-	char *res = NULL, *lnk = NULL;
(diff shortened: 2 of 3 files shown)


jnbr:
Nice!
The test looks strange to me. Shouldn't this all be `pkg_A` ?

xtraeme:
I think it didn't matter, the key is to have some dangling symlinks. Will
review it later.

Thanks

xtraeme:
Yeah, just checked it. It does not matter, the test is fine.

xtraeme:
ping?

jnbr:
>  the key is to have some dangling symlinks

I don't think you get dangling symlinks this way. 
You need a non existent target for a dangling symlink but the path of the linkname should exist which isn't the case when using `pkg_B` in this test:

```
$ mkdir -p pkg_A/usr/lib
$ touch -f pkg_A/usr/lib/libfoo.so.1.2.0
$ ln -sf libfoo.so.2 pkg_A/usr/lib/libfoo.so.1
$ tree
.
└── pkg_A
    └── usr
        └── lib
            ├── libfoo.so.1 -> libfoo.so.2
            └── libfoo.so.1.2.0

3 directories, 2 files
$ ln -sf libfoo.so.2 pkg_B/usr/lib/libfoo.so
ln: failed to create symbolic link 'pkg_B/usr/lib/libfoo.so': No such file or directory
$ tree
.
└── pkg_A
    └── usr
        └── lib
            ├── libfoo.so.1 -> libfoo.so.2
            └── libfoo.so.1.2.0

3 directories, 2 files
```

xtraeme:
@jnbr you were right, I changed it to pkg_A, as expected.

xtraeme:
This PR is ready for a long while... real case: fixes complete removal of faenza-icon-theme and probably others containing multiple levels of symlinks.

Would be nice if it's applied...

PR #95, opened 2019-05-18
libxbps: refactor default and relative path code

CameronNemo:


lib/initend.c
```diff
@@ -407,14 +407,39 @@ parse_dir(struct xbps_handle *xhp, const char *cwd, const char *confdir, const c
 	return rv;
 }
 
+static int
+fallback_path(char *dest, size_t size, char *root, const char *fallback)
+{
+	int rv;
+	rv = snprintf(dest, size, "%s%s", strncmp(root, "/", 2) ? root : "",
+	              fallback);
+	if (rv < 0 || (size_t)rv >= size)
```
Duncaen:
```suggestion
return rv < 0 || (size_t)rv >= size;
```
But maybe it would be better to return `ENOMEM`, 1 is `EPERM` which is kinda confusing.

CameronNemo: Changed to return ENOMEM rather than 1.

xtraeme:
any progress on this? I can merge this manually but I'd like to merge the PR instead :-)

xtraeme:
superseded by #222

PR #101, opened 2019-06-11
xbps alternatives: multiple fixes, new test cases.

xtraeme:
In the !current branch, "first" wasn't initialized so it displayed garbage:

```
Removing `pinentry-tty-1.1.0_5' ...
Removing 'pinentry' alternatives group symlink: pinentry
pinentry-tty-1.1.0_5: unregistered 'pinentry' alternatives group
Switched 'pinentry' alternatives group to 'p
o@�'
Creating 'pinentry' alternatives group symlink: pinentry -> /usr/bin/pinentry-tty
Removed file `/usr/bin/pinentry-tty'
Removed `pinentry-tty-1.1.0_5' successfully.
```

With my fix:
```
Removing `pinentry-tty-1.1.0_5' ...
Removing 'pinentry' alternatives group symlink: pinentry
pinentry-tty-1.1.0_5: unregistered 'pinentry' alternatives group
Switched 'pinentry' alternatives group to 'pinentry'
Creating 'pinentry' alternatives group symlink: pinentry -> /usr/bin/pinentry-tty
Removed file `/usr/bin/pinentry-tty'
Removed `pinentry-tty-1.1.0_5' successfully.
```

Signed-off-by: Juan RP <xtraeme@gmail.com>

### DIFF
--- a/lib/package_alternatives.c
+++ b/lib/package_alternatives.c
@@ -366,10 +366,10 @@ xbps_alternatives_unregister(struct xbps_handle *xhp, xbps_dictionary_t pkgd)
 		if (!update && !current)
 			continue;
 
+		xbps_array_get_cstring_nocopy(array, 0, &first);
+
 		if (!current) {
 			/* get the new alternative group package */
-			first = NULL;
-			xbps_array_get_cstring_nocopy(array, 0, &first);
 			curpkgd = xbps_pkgdb_get_pkg(xhp, first);
 			assert(curpkgd);
 		}
--- a/lib/package_alternatives.c
+++ b/lib/package_alternatives.c
@@ -492,8 +492,7 @@ xbps_alternatives_register(struct xbps_handle *xhp, xbps_dictionary_t pkg_repod)
 			rv = create_symlinks(xhp,
 				xbps_dictionary_get(pkg_alternatives, keyname),
 				keyname);
-			if (alloc)
-				xbps_object_release(array);
+			xbps_object_release(array);
 			if (rv != 0)
 				break;
 		}
--- a/lib/package_alternatives.c
+++ b/lib/package_alternatives.c
@@ -355,8 +355,10 @@ xbps_alternatives_unregister(struct xbps_handle *xhp, xbps_dictionary_t pkgd)
 
 		xbps_set_cb_state(xhp, XBPS_STATE_ALTGROUP_REMOVED, 0, NULL,
 		    "%s: unregistered '%s' alternatives group", pkgver, keyname);
-		if (!update)
+		if (!update) {
 			xbps_remove_string_from_array(array, pkgname);
+			xbps_array_get_cstring_nocopy(array, 0, &first);
+		}
 
 		if (xbps_array_count(array) == 0) {
 			xbps_dictionary_remove(alternatives, keyname);
@@ -366,9 +368,7 @@ xbps_alternatives_unregister(struct xbps_handle *xhp, xbps_dictionary_t pkgd)
 		if (!update && !current)
 			continue;
 
-		xbps_array_get_cstring_nocopy(array, 0, &first);
-
-		if (!current) {
+		if (current) {
 			/* get the new alternative group package */
 			curpkgd = xbps_pkgdb_get_pkg(xhp, first);
 			assert(curpkgd);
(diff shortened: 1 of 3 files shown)


xtraeme:
Hmm looking at this output, something seems completely wrong

xtraeme:
I removed pinentry-tty, `pinentry` symlink should now point to `pinentry-curses`, but it didn't.

So I suspect this code is broken in this case.

Duncaen:
http://sprunge.us/dKjF8Y

If the package is the current alternative, we need to use the next new alternative.
But I somehow think there is another logic error.

Duncaen:
I think the missing step is if `current == true` and `update == true`, but the updated package doesn't provide the alternative group in question anymore.

xtraeme:
Got it, hold on.

Duncaen:
Here is another test case http://sprunge.us/bnhe1F

xtraeme:
This test case fails (less_entries_update) currently. Looking at it.

Duncaen:
Its getting the old packages metadata when using `xbps_pkgdb_get_pkg`, so it will remove the symlinks and just recreate the same again.

xtraeme:
In your new test case what's happening is that alternatives from old version need to be unregistered before registering the ones from the updated pkg.

xtraeme:
Ok, found a way to fix this. Hold on.

xtraeme:
@Duncaen fixed.

Duncaen:
Hm, but the same happens in the other direction too.

```
atf_test_case more_entries_update

more_entries_update_head() {
    atf_set "descr" "xbps-alternatives: add symlinks provided by package update"
}
more_entries_update_body() {
    mkdir -p repo pkg_A/usr/bin pkg_B/usr/bin
    touch pkg_A/usr/bin/A1 pkg_A/usr/bin/A2 pkg_B/usr/bin/B1

    cd repo
    xbps-create -A noarch -n A-1.1_1 -s "A pkg" --alternatives "1:1:/usr/bin/A1" ../pkg_A
    atf_check_equal $? 0
    xbps-rindex -d -a $PWD/*.xbps
    atf_check_equal $? 0
    cd ..

    xbps-install -r root --repository=repo -ydv A
    atf_check_equal $? 0

    cd repo
    xbps-create -A noarch -n A-1.2_1 -s "A pkg" --alternatives "1:1:/usr/bin/A1 1:2:/usr/bin/A2" ../pkg_A
    atf_check_equal $? 0
    xbps-rindex -d -a $PWD/*.xbps
    atf_check_equal $? 0
    cd ..

    xbps-install -r root --repository=repo -ydvu
    atf_check_equal $? 0

    rv=0
    [ -e root/usr/bin/2 ] || rv=1
    atf_check_equal $rv 0
}
```

xtraeme:
@Duncaen looking at it, I added this test case and fails.

Duncaen:
Also changing symlink targets would fail currently.
I generally think it might be better to do the alternatives switch later i.e. in `unpack_archive` where we have `binpkg_propsd` and can rely on the data being correct because its signed.

xtraeme:
Test cases: 181 total, 0 skipped, 0 expected failures, 0 broken, 0 failed

xtraeme:
@Duncaen any other test case you want to add?

Duncaen:
Ah this looks good, I guess that also fixed changed symlinks in updates. I don't have a testcase yet because I have to leave in a few minutes but I can write one later.

lib/package_alternatives.c
```diff
@@ -438,8 +472,15 @@ xbps_alternatives_register(struct xbps_handle *xhp, xbps_dictionary_t pkgd)
...
-				continue;
+			if (xbps_match_string_in_array(array, pkgname)) {
+				/* apply alternatives for this group */
+				rv = create_symlinks(xhp,
+					xbps_dictionary_get(pkg_alternatives, keyname),
+					keyname);
+				if (rv != 0)
+					break;
+			}
+			continue;
```
huglovefan: the `if (alloc)` later on is now always true since this branch always continues

xtraeme: Fixed, thanks.

xtraeme:
Hold on, found another issue.

xtraeme:
Ok, everything fixed. It works well enough now for me.

This branch is ready to be merged, let me know if you've found any other issue.

xtraeme:
From travis:

```
libgcc-devel-8.3.0_1.x86_64.xbps: [1274KB 67%] 49KB/s ETA: 00m08s
```

seriously? omg

PR #112, opened 2019-06-13
xbps-checkvers

Duncaen:


bin/xbps-checkvers/main.c
```diff
@@ -200,9 +200,10 @@ show_usage(const char *prog)
 "  -I,--installed 		Check for outdated packages in rootdir, rather\n"
 "				than in the XBPS repositories.\n"
 "  -i,--ignore-conf-repos	Ignore repositories defined in xbps.d.\n"
+"  -m,--manual	Only process listed files.\n"
```
xtraeme: Please fix the indent here.

xtraeme:
so that with -m we can now specify a list of templates as arguments?

xtraeme:
Manual page needs to be updated ?

xtraeme:
oh I see, you've combined -s,--show-all with -m.

Great

xtraeme:
:+1:

bin/xbps-checkvers/main.c
Duncaen: Fixed it while making the usage text match all the other xbps-* usage outputs.

PR #115, opened 2019-06-14
xbps-checkvers --format

Duncaen:
from the man page:

```
     -f, --format format
         Format according to the string format, inspired by printf(3).

         The following formatting codes may be used:

         \n      Newline.

         \t      Tab.

         \0      NULL.

         %%      A plain ‘%’.

         %n      The package name.

         %r      The repository version.

         %s      The source package version.

         The default format is “%n %r %s”.
```

The other commit makes it easier to just pass a list of package names, instead of having to construct paths.

### DIFF
--- a/bin/xbps-checkvers/main.c
+++ b/bin/xbps-checkvers/main.c
@@ -635,8 +635,7 @@ static void
 rcv_printf(rcv_t *rcv, FILE *fp, const char *pkgname, const char *repover,
     const char *srcver)
 {
-	char tmpl[128], *p;
-	const char *f;
+	const char *f, *p;
 
 	for (f = rcv->format; *f; f++) {
 		if (*f == '\\') {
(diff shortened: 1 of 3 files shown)


bin/xbps-checkvers/main.c
```diff
@@ -628,12 +631,53 @@ check_reverts(const char *repover, const map_item_t reverts)
 	return rv;
 }
 
+static void
+rcv_printf(rcv_t *rcv, FILE *fp, const char *pkgname, const char *repover,
+    const char *srcver)
+{
+	char tmpl[128], *p;
```
xtraeme: 128 is a magic value? :-)

bin/xbps-checkvers/main.c
```diff
@@ -591,18 +476,17 @@ rcv_set_distdir(rcv_t *rcv, const char *distdir)
...
-		return rv;
+	assert(reverts);
+	if (!(len = strlen(reverts)))
+		return false;
 
-	sreverts = calloc(reverts.v.len+1, sizeof(char));
-	assert(sreverts);
-	strncpy(sreverts, reverts.v.s, reverts.v.len);
-	sreverts[reverts.v.len] = '\0';
+	assert((sreverts = strdup(reverts)));
```
xtraeme: I would change this to not use assert(), because there's an alloc that could be missed if compiled with -DNDEBUG

Issue #3, opened 2018-06-28
xbps-install -Mun does take to long when one repo is not reachable


Johnnynator:
`xbps-install -Mun` does hang in an infinite loop when at least one repos is not reachable, which makes `xuname` hanging forever.
Weirdly this does only happen when using xbps that was compiled with `xbps-src`. Compilling it manually and running `bin/xbps-install/xbps-install -Mun` does somehow work.

strace looping part
```
socket(AF_INET, SOCK_STREAM, IPPROTO_TCP) = 4
connect(4, {sa_family=AF_INET, sin_port=htons(9090), in_addr=inet_addr("192.168.178.49")}, 16) = -1 EHOSTUNREACH (Host is unreachable)
close(4)
```

Gottox:
@Johnnynator Do you use master to compile?

Johnnynator:
I tested it with daa1b446821646d8923623f8a0d6e1b65db0929b . Which shold be the same as the xbps package + the patches.

Johnnynator:
Just retested and I now run always into that bug. idk why it did work in the first try then

Gottox:
Do you use a proxy?

Johnnynator:
Nope. Network does not have any special configuration.
I just have a local repo for some stuff which is not always active.
`xbps-install -u` is working fine while `xbps-install -Mu` will not work.

Johnnynator:
Actually the problem is probably my expectation at how `-M` works compared to `-S`. `-S` seems to skip everything that is not reachable while `-M` retires to get the repo.

Gottox:
I'm able to reproduce the issue with `xbps-install -Mun --repository=http://100.0.0.1:9001`

Duncaen:
Its actually not an infinity loop.

For each installed package it iterates over the repository pool with `xbps_rpool_foreach`.
In `xbps_rpool_foreach` it calls `xbps_repo_open`, which if `XBPS_FLAG_REPOS_MEMSYNC` requests the repo data each time.

This results in `${number_installed_packages} * ${number_of_repositories}` http requests.

xtraeme:
I'd name this a bug rather than enhancement.

xtraeme:
This should be fixed now.

Johnnynator:
Yep this works fine now.

PR #120, opened 2019-06-15
lib/repo: add xbps_repo_remove and use it in xbps_rpool_foreach

Duncaen:
https://github.com/void-linux/xbps/issues/3


xtraeme:
Looks great buddy :+1:

PR #121, opened 2019-06-16
Transaction files

Duncaen:
```
duncaen | ok, its fast enough when we store the file size in files.plist and only read the archive until files.plist is found, which is always within the first 4 entries
duncaen | This should fix xbps deleting wrong files if a package is split
duncaen | and adding the disk space check will be easy too
duncaen | and files changing types should be easier to handle too
```


Duncaen:
Going to rework it a bit to remove obsolete files for each package before unpacking to fix the issues replacing directories/symlinks with files.
Adding the test from https://github.com/void-linux/xbps/pull/22 and another test I have for replacing symlinks with files.

xtraeme:
:+1:

Duncaen:
@xtraeme so I run into this test: https://github.com/void-linux/xbps/blob/master/tests/xbps/libxbps/shell/issue31_test.sh, with the new behavior xbps deletes the files.
Can I update the test behavior?
Or should I create another untracked file in the directory that gets deleted to make `xbps-install` return `ENOTEMPTY` again?

xtraeme:
@Duncaen yes, please update the test case.

Duncaen:
Ok this will be good, I can easily check if a directory that needs to be removed can be deleted, by counting the entries.
So if there is a directory with an obsolete file with hash mismatch or some untracked file, xbps can abort before the transaction even starts.

xtraeme:
This is great!

xtraeme:
I think I documented all my changes in the `NEWS` file.
@Duncaen please do this asap, once this feature is pushed it will be locked with only bugfixes for some days, then we will release 0.54.

lib/transaction_files.c
```diff
@@ -0,0 +1,446 @@
...
+		return ENOMEM;
+	}
+
+	/*
+	 * Enable support for tar format and gzip/bzip2/lzma compression methods.
+	 */
+	archive_read_support_compression_gzip(ar);
+	archive_read_support_compression_bzip2(ar);
+	archive_read_support_compression_xz(ar);
+	archive_read_support_format_tar(ar);
```
xtraeme: I think the number of times we do this is embarrassing. We'd need a private function to remove redundant code.

Also please change them to `archive_read_filter_compression_xx`, the `archive_read_support_xxx` are from the old interface.

Duncaen:
I think I'm done, need to test it a bit and maybe add some more test cases that show that transactions are aborted before they even start.

xtraeme:
great! can't wait to see how this works in real time :-)

lib/package_unpack.c
```diff
@@ -111,6 +110,31 @@ unpack_archive(struct xbps_handle *xhp,
...
+	/*
+	 * Remove obsolete files.
+	 */
+	if (!preserve &&
+	    xbps_dictionary_get_dict(xhp->transd, "obsolete_files", &obsd) &&
+	    (obsoletes = xbps_dictionary_get(obsd, pkgname))) {
+		for (unsigned int i = 0; i < xbps_array_count(obsoletes); i++) {
+			const char *file = NULL;
+			xbps_array_get_cstring_nocopy(obsoletes, i, &file);
+			fprintf(stderr, ">>> %s\n", file);
```
xtraeme: Please remove this one or change it to xbps_dbg_printf()

Duncaen:
What should I do about `xbps_find_pkg_obsoletes`. It is a public API should I just remove it or might there be any need for it in the future?

xtraeme:
We've bumped major recently, if it's not necessary anymore just remove it and bump api date in xbps.h

tests/xbps/libxbps/shell/obsoletefiles_test.sh
```diff
@@ -458,6 +458,31 @@ nonempty_dir_abort_body() {
 	atf_check_equal $? 0
 }
 
+atf_test_case conflicting_files_in_transaction
+
+conflictinc_files_in_transaction_head() {
```
xtraeme: typo

xtraeme:
is it ready now?

Duncaen:
I think so, but I feel like there are not enough test using `xbps-pkgdb -a` at the end.

xtraeme:
We can improve this later. We will have some days for testing.

PR #130, opened 2019-06-20
Obsoletes preserve

Duncaen:
Test case 1. where a directory with preserved files is changed to another file type, will abort before transaction.
Test case 2. where a preserved symlink is replaced with a file, will just skip considering it obsolete and leave it as it is.

### DIFF
--- a/tests/xbps/libxbps/shell/obsoletefiles_test.sh
+++ b/tests/xbps/libxbps/shell/obsoletefiles_test.sh
@@ -136,11 +136,11 @@ files_move_from_dependency_body() {
 	xbps-install -r root -C null.conf --repository=$PWD -yuvd
 	atf_check_equal $? 0
 
-	foofile=$(xbps-query -r root -f A-1.1_1|grep foo)
-	atf_check_equal $foofile ""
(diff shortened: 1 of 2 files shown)


tests/xbps/libxbps/shell/obsoletefiles_test.sh
```diff
@@ -326,6 +326,88 @@ directory_to_symlink_body() {
...
+	echo "fail" >../pkg_A/foo/bar
+	xbps-create -A noarch -n A-1.1_1 -s "A pkg" ../pkg_A
+	atf_check_equal $? 0
+	xbps-rindex -d -a $PWD/*.xbps
+	atf_check_equal $? 0
+	cd ..
+
+	xbps-install -r root -C $PWD/xbps.d --repository=$PWD/some_repo -dvyu
+	# EEXIST
+	atf_check_equal $? 0
```
xtraeme: no EEXIST

Duncaen: Right, removed the comment.

PR #135, opened 2019-06-22
lib/fetch: try all addresses before aborting

jnbr:


jnbr:
doesn't work:
```
strace -e connect /usr/bin/xbps-fetch -v https://schlomp.space/tastytea/hashboot/releases/download/0.9.14/hashboot-0.9.14.tar.gz
looking up schlomp.space
connect(3, {sa_family=AF_INET, sin_port=htons(65535), sin_addr=inet_addr("79.195.155.162")}, 16) = 0
connect(3, {sa_family=AF_INET6, sin6_port=htons(65535), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "2003:c0:870c:6600::1:1", &sin6_addr), sin6_scope_id=0}, 28) = 0
connecting to schlomp.space:443
connecting to 2003:c0:870c:6600::1:1
connect(3, {sa_family=AF_INET6, sin6_port=htons(443), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "2003:c0:870c:6600::1:1", &sin6_addr), sin6_scope_id=0}, 28) = -1 EINPROGRESS (Operation in progress)
https://schlomp.space/tastytea/hashboot/releases/download/0.9.14/hashboot-0.9.14.tar.gz: Connection refused
+++ exited with 1 +++
```

jnbr:
should work now but needs more testing

Duncaen:
I think this would need `waiting -= n6-i6` when `ENETUNREACH` otherwise it would never hit 0, not sure if this is even necessary.

Duncaen:
Or maybe move `if (!waiting) break;` before the `poll(2)` and keep waiting as variable to count waiting sockets.

jnbr:
> Or maybe move if (!waiting) break; before the poll(2) and keep waiting as variable to count waiting sockets.

I did that.

lib/fetch/common.c
```diff
@@ -639,8 +641,6 @@ happy_eyeballs_connect(struct addrinfo *res0, int verbose)
 			rv = -1;
 			break;
 		}
-		if (!waiting)
-			break;
 		else if (done)
```
Duncaen: I think this can be removed, this is another bug.
If it tried all addresses done is 1. When there are >1 sockets it is waiting for, in the last long 300 second timeout and then one of the >1 sockets fails to connect i.e. with ECONNREFUSED, it will just stop waiting for the other sockets that may be established later.

jnbr: It makes a difference if all addresses fail to connect. In this case, without the check it would wait for the timeout while with the check, it fails immediately.
I think this fast failure is useful.

Duncaen: But as is is, its dead code. It needs to be `if (done)`.

Not sure if fast failure is good. If there are two addresses, one ipv4 and one ipv6 and they are both too slow to respond within the connection delay, it will set done=1 and then set the timeout to 300 seconds.
If now one of them fails the other one would be closed.

jnbr: oops, I was referring to the `if(!waiting)` which I moved on top.

Duncaen:
lgtm

PR #136, opened 2019-06-22
lib/plist.c: fix xbps_array_foreach_cb_multi processing entries multiple times

Duncaen:
the first thread to finish will start again from 0 (reserved) until 0+slicecount,
next thread will then start from the end of the previous thread which
is already wrong, resulting in processing the first slicecount*maxthreads
entries twice.

The first slicecount*maxthreads entries are accounted by the thread
creation loop and `reserved` has to start at the first unaccounted index.

### DIFF
--- a/lib/plist.c
+++ b/lib/plist.c
@@ -111,11 +111,9 @@ xbps_array_foreach_cb_multi(struct xbps_handle *xhp,
 	if (xbps_object_type(array) != XBPS_TYPE_ARRAY)
 		return 0;
 
-	if (!xbps_array_count(array))
+	arraycount = xbps_array_count(array);
+	if (arraycount == 0)
 		return 0;
-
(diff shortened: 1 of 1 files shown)


lib/plist.c
```diff
@@ -138,6 +136,8 @@ xbps_array_foreach_cb_multi(struct xbps_handle *xhp,
 		}
 	}
 
+	reserved = slicecount * maxthreads;
```
Gottox: make sure reserved isn't initialized beforehand. This way we get a compile time warning when it's used before it's set to the correct value

Duncaen: Done, not sure if it actually gives a warning, because the pointer is passed arround.

xtraeme:
Works nicely, well done!

PR #149, opened 2019-07-07
Transaction fetch & verify

Duncaen:
This changeset will digest the file while its being downloaded and verify it when the download is finished.


Anachron:
No test cases to add?

Duncaen:
There are no test cases for any fetch functionality, because we would have to embed a webserver.
This something for another PR.

PR #162, opened 2019-08-05
xbps-install: remove obsolete packages during update

Chocimier:
This is proof of concept phase of automatic removal of obsolete installed packages, with optionally installing other ones in exchange.
This is meant to solve two problems: explicit removal of packages removed from repo, like https://github.com/void-linux/void-packages/pull/13423 and automatic replacing packages, like https://github.com/void-linux/void-packages/pull/13379 . This PR is not meant to unindex packages to not be able to install them.

Code is based on xbps_transaction_package_replace.

Open problems:
- **providing data for xbps-install**
Hard-coded list is not good enough. The list should be included into repodata to make sure it is synchronized with available packages. There are two possible ways to do that: add into root element of index.plist new top-level a metadata key that will contain the list, or add new file in addition to index.plist and index-meta.plist.
- **source of data for xbps-rindex**
For packages to be removed, we want to completely remove templates, so there is no place to put information about what packages needs to be removed from installations. New common/successions file may be added to void-packages repository from which list in repodata will be generated.
- **virtual packages**
I need some explanation why they need to be processed differently than regular packages.
- **decide which packages to remove**
Packages turned into removed packages as per current Manual.md should be automatically removed. For other, that were removed because of e.g. not able to rebuild but otherwise were OK and were removed by dropping template (and possibly removing xbps from mirror), individual decision could be made by generally they can stay installed.
- **not creating empty transactions**
Just a detail that will be solved with return code.
- **efficiency**
- **terminology**: succession, retirement, obsolescence, …

Lot of tests need to be done, e.g. automatic-install copying, virtual packages, interference with updates etc.

Waiting for opinions.


Duncaen:
One problem with repodata as source is, that its not signed. Most other metadata is at least checked against what the signed binary package provides on install/update.

Chocimier:
Is there a reason to not put signature of plists into repodata?

Vaelatern:
The reason we have is that signatures are made at a different time than the repodata is updated.

Chocimier:
Ok, but why not to leave signing packages as is and start to sign repodata (it is xbps-rindex too)?
Presence of privkey during build is considered a security risk, I guess?

Or if it is not an option, do you see some other way to publish obsoletes list?

Vaelatern:
Signing repodata would be good.

Chocimier:
Signing index.plist for now, but I think it will be easier to add third plist and modify reading archive than to modify iterating packages in index.plist to skip metadata.

Chocimier:
index.plist is signed now. There is other place (xbps_repo_fetch_remote), where signature should be verified.

Chocimier:
Now I only see minor things to change, like passing to xbps-rindex simpler data format than XML.

Please review.

Chocimier:
To sum up progress:
- providing data for xbps-install: signed index-meta.plist
- source of data for xbps-rindex: from [text file similar to shlibs](https://github.com/Chocimier/xbps-org/blob/succession/tests/xbps/xbps-install/data/outmoded), passed with argument
- virtual packages: not tested as I do not see use case
- decide which packages to remove: to be done if change accepted
- not creating empty transactions: done
- efficiency: ok
- terminology: chosen "outmoded"
- tests: it works, some kyua tests written

xtraeme:
Could you please move the signed index-meta.plist changes to another PR?

xtraeme:
is this still relevant?

Chocimier:
Relevant if we want to change uninstalling by revbump to something predictable.
I do, but do not know how to broadcast list of packages to remove without signed repodata.

Chocimier:
As discussed on irc, we should rather remove packages missing from repo without keeping list.

PR #164, opened 2019-08-13
lib/package_script.c: try to find a usable shell instead of using /bin/sh

Gottox:


xtraeme:
Looks ok, can you restart the travis build again? just to check.

xtraeme:
I would expand this to cover other common shells: zsh and ksh. As we use POSIX shell scripts any shell of them would work.

```#!/usr/bin/env ...``` maybe?

**Comment by Gottox** [comment]
Is zsh posix compliant?

**Comment by xtraeme** [comment]
iirc it does.

**Comment by Gottox** [comment]
tbh, the more I think about it, the worse I find this workaround. We never should remove `sh` in the first place.

**Comment by xtraeme** [comment]
Yeah... alternatives should use rename() for a proper fix

**Comment by Gottox** [comment]
instead of link()?

**Comment by xtraeme** [comment]
from rename(2) from man-pages:

```
      If oldpath refers to a symbolic link, the link is renamed; if newpath
       refers to a symbolic link, the link will be overwritten.
```

more or less like proplib creating the plist files. Make the symlink with a tempname, on success rename()

Duncaen:
The problem is that the symlinks are removed from `xbps_remove_pkg` and then registered again after unpacking.
But before unpacking it runs the pre-install scriptlet and will fail.

I think logically this is the correct way to go, but I guess we could just not call `xbps_alternatives_unregister` from `xbps_remove_pkg` if its an update.

Duncaen:
There are also other issues with alternatives, like with the current youtube-dl update, where the package providing the alternative becomes a meta package without alternatives depending on a package that provides the alternative as regular file. 
So unpacking the new package will overwrite the old alternatives symlink with the regular file and then afterwards when the previous alternative package is update to the meta package it will unlink the alternative and remove the regular file.
I added a patch to the live xbps to only unlink if its a symlink, but I think that it is not the correct solution and it would still remove the wrong file if the alternative was switched to not being an alternative but still is a symlink.

xtraeme:
how about this one? seems to pass all tests... http://dpaste.com/2GKZ8QK

Duncaen:
Should work, but I think it would be better to get rid of the `alternatives-update` special case and just don't call `xbps_alternatives_unregister` instead.

xtraeme:
Either way is ok to me

Duncaen:
Also as addition, what do you think about making the scriptles if they exit nonzero not abort the transaction if they fail and maybe mark them as unconfigured?
I think installing half the transaction is a worse result than maybe missing some pre/post-install script runs.

xtraeme:
Sounds good

xtraeme:
Ok, I fixed this for you to make travis succeed.

PR #173, opened 2019-10-25
Archive write errors

Duncaen:
This resolves the issues we've seen when `/hostdir` is full.

1. `xbps-rindex` would overwrite repodata with an empty file if some of the writes fail, in our case we lost some staged packages from the index. 
1. `xbps-create` would create incomplete archives that can't be added to the repository index and would also make `xbps-src` skip registering a new build until all corrupt `.xbps` files were removed manually.


xtraeme:
👌

PR #174, opened 2019-10-28
Signing repo metadata

Chocimier:
Signing extracted from #162. No tests yet.

cc @xtraeme


lib/repo.c
```diff
@@ -77,7 +114,44 @@ repo_get_dict(struct xbps_repo *repo)
...
+}
+
+
+xbps_dictionary_t HIDDEN
+get_safe_idxmeta(xbps_dictionary_t full) {
+	static const char *keys[] = {
+		"public-key",
+		"public-key-size",
+		"signature-by",
+		"signature-type",
```
CameronNemo: What do these fields represent?

Chocimier: They describe which key in `/var/db/xbps/keys/` should be used to verify binary packages.

Chocimier:
@xtraeme please review.

PR #176, opened 2019-11-02
xbps_verify_digest_signature: use "public-key" obj from idxmeta.

xtraeme:
... rather than trying to read it via rootdir/metadir/hexfp.plist.

The 'idxmeta' dictionary contains all objs already, so let's use it.

This fixes the xbps-rindex/sign testcases.


xtraeme:
@Chocimier xbps-rindex now does not need -r (rootdir).

Chocimier:
Verifying message with key attached to this message is pointless. Security require that key is send with another channel than message.

xtraeme:
I'll fix this with another approach.

PR #177, opened 2019-11-04
xbps-{uunshare,uchroot}: restrict option parsing

sgn:
GNU's getopt(3) and getopt_long(3) allows options anywhere in the
command. Thus, crippling (-b /tmp:/tmp is used for cosmetic, but not
required to demonstrate the problem):

    xbps-uunshare -b /tmp:/tmp /masterdir sh -c 'echo yes'

We could use:

    xbps-uunshare -b /tmp:/tmp -- /masterdir sh -c 'echo yes'

to overcome that problem but it isn't documented in our man-pages.

Consider the limited usage of this utility, just stop parsing options
when we get our first non-options argument.

Signed-off-by: Doan Tran Cong Danh <congdanhqx@gmail.com>

### DIFF
--- a/bin/xbps-uchroot/main.c
+++ b/bin/xbps-uchroot/main.c
@@ -281,7 +281,7 @@ main(int argc, char **argv)
 	tmpfs_opts = chrootdir = cmd = NULL;
 	argv0 = argv[0];
 
-	while ((c = getopt_long(argc, argv, "Oto:b:V", longopts, NULL)) != -1) {
+	while ((c = getopt_long(argc, argv, "+Oto:b:V", longopts, NULL)) != -1) {
 		switch (c) {
 		case 'O':
 			overlayfs = true;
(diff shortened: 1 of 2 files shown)


Duncaen:
I did this before, but it was reverted: https://github.com/void-linux/xbps/commit/449dc8a5517f870d52f0cef9daf7fa69aefeff0e

zdtcd:
On 2019-11-04 08:11:09 -0800, Duncan Overbruck wrote:
> I did this before, but it was reverted:
> https://github.com/void-linux/xbps/commit/449dc8a5517f870d52f0cef9daf7fa69aefeff0e

I should update the documentation instead then.

-- 
Danh

zdtcd:
I guess the change broke someone setup that depends on free positioning.

xtraeme:
Not relevant anymore with #178 and it was reverted. Anyway would be nice to know why.

PR #179, opened 2019-11-14
lib/package_alternatives.c: prune obsolete alternatives groups

q66:
In the edge case when an updated package has different (or no) alternatives groups, make sure to prune those that are in pkgdb but not in the newly installed package.

A potentially common case of this is when a package that formerly had alternatives gets removed and a transitional metapackage takes its place (which has no alternatives).

I ran across this when updating `openjdk` and renaming it to `openjdk8` and adding transitional packages. I don't think it can be updated before a fix for this is in xbps, either this one or something better...

### DIFF
--- a/lib/package_alternatives.c
+++ b/lib/package_alternatives.c
@@ -309,6 +309,22 @@ xbps_alternatives_set(struct xbps_handle *xhp, const char *pkgname,
 	return rv;
 }
 
+static int
+switch_alt_group(struct xbps_handle *xhp, const char *grpn, const char *pkgn,
+		xbps_dictionary_t *pkg_alternatives)
+{
+	xbps_dictionary_t curpkgd, pkgalts;
+
+	curpkgd = xbps_pkgdb_get_pkg(xhp, pkgn);
+	assert(curpkgd);
+
+	xbps_set_cb_state(xhp, XBPS_STATE_ALTGROUP_SWITCHED, 0, NULL,
+		"Switched '%s' alternatives group to '%s'", grpn, pkgn);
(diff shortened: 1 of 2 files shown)


q66:
I attempted to come up with slightly more elaborate behavior than my first try.

Basically, the intended/implemented behavior is:

```
if removed:
    remove provided alternatives
else if installed:
    set up provided alternatives
else if upgraded:
    if old alternatives != new alternatives:
        prune obsolete symlinks
        prune obsolete groups
    if currently set alternatives != from this package:
        do nothing
    else if num_deps(installed_pkg) == 0:
        assume removed package in xbps-src
        switch to first available alternative groups (same as normal removal)
    else:
        assume renamed package in xbps-src (transitional)
        switch to last available alternative groups (most recently installed)
```

I updated the tests to include checking that behavior also.

q66:
I kinda wonder if it'd be worth it to try something even more elaborate, say, inspect the dependencies of the upgraded package and only switch to "newest" alternatives group if it's provided by a package in the dependencies list.

But OTOH that would result in more complicated code, with potentially error-prone checking of questionable value, and I can think of more edge cases breaking it... so I think this should be fine, at very least it's predictable and consistent in its behavior (in the very worst case, which is unlikely to happen ever as it would require 3+ packages providing the same alternatives group at the same time, e.g. `A B C`, it would switch to C instead of B... this would happen should a package *replace* its alternatives set with a completely different one, while still having dependencies - i.e. not being meta, and then two other foreign packages providing the previous alternatives set)

Thoughts, @xtraeme @Duncaen ?

xtraeme:
looks ok to me. Could you please rebase on master and force push? it should fix the stupid error seen with xbps.

PR #181, opened 2019-11-24
add test which replaces an alternative group with a symlink

jnbr:
This bug was triggered by removing the cc alternative group from gcc and clang.
I forgot to this variant with the latest xbps.

### DIFF
--- a/tests/xbps/xbps-alternatives/main_test.sh
+++ b/tests/xbps/xbps-alternatives/main_test.sh
@@ -766,6 +766,40 @@ prune_leftover_groups_body() {
 	atf_check_equal $rv 0
 }
 
+atf_test_case replace_alternative_with_symlink
+
+replace_alternative_with_symlink_head() {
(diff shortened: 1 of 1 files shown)


xtraeme:
right, this fails currently... do you have a fix?

jnbr:
no, not yet

Duncaen:
We worked around this with https://github.com/void-linux/xbps/commit/98e09e01e1d254347ab579ba849aefcb640ff8f1 for normal files and directories.

I think the only proper fix is to use the transactional files stuff, because even if we would change to updates alternatives before unpacking a package, it would still fail if the alternative was removed and moved to another package which was previously unpacked in the transaction.

xtraeme:
FTR I just fixed this one with https://github.com/void-linux/xbps/commit/732a0e6bdf6a59e761d72fda291d52038358fdfe

jnbr:
Thanks!

PR #183, opened 2019-11-28
lib/verifysig.c: don't read keys from cwd

Chocimier:


xtraeme:
Why xbps-rindex needs to know about the rootdir? imho it's not the correct way to fix this.

Chocimier:
Apart from testing, it may not need. But reading keys from cwd by all xbps including xbps-install isn't correct way either.

xtraeme:
Manpage needs to be updated

Chocimier:
Already documented in xbps-rindex manpage. There is long option only.

Duncaen:
Imho we should just revert everything for now, we can't use the repodata signing code as it is anyways.

Chocimier:
> Imho we should just revert everything for now, we can't use the repodata signing code as it is anyways.

Fine for me.

xtraeme:
shall we revert the repodata signing stuff? I don't really like the fact that xbps-rindex now needs to know about rootdir, which to me is wrong.

xtraeme:
the repodata signing patchset has been reverted, feel free to submit again.

PR #185, opened 2019-11-30
tests: replace file with alternative

CameronNemo:
trying to emulate https://github.com/void-linux/void-packages/pull/15836


CameronNemo:
this is a blocker for busybox base system, as we cannot switch binaries (like ip, tc, hwclock, etc) to an alternative.

xtraeme:
I merged this test case, thanks.

Issue #157, opened 2019-07-17
Unnecessary prompt for updating packages on hold


svenper:
If the only packages that would be affected by an update are on hold, does it make any difference whether you answer yes or no? Prompting seems unnecessary to me, since no packages will be changed.

```text
# xbps-install -Su 
[*] Updating `https://alpha.de.repo.voidlinux.org/current/musl/x86_64-musl-repodata' ...
x86_64-musl-repodata: 1796KB [avg rate: 6104KB/s]
[*] Updating `https://alpha.de.repo.voidlinux.org/current/musl/nonfree/x86_64-musl-repodata' ...

Name        Action    Version           New version            Download size
linux       hold      4.14_1            4.19_2                 -

Space available on disk:        212GB

Do you want to continue? [Y/n]
```

xtraeme:
fixed with #188

PR #188, opened 2019-12-10
xbps-install: early exit when all packages on hold (no prompt)

mbarbar:
Lovely codebase.

This PR is for issue #157 .

Simply, checks if all packages are on hold and exits if so. This happens immediately before the [Y/n] prompt appears, so after the actions/sizes are shown. I think this is more appropriate than before the actions/sizes are shown.

### DIFF
--- a/bin/xbps-install/transaction.c
+++ b/bin/xbps-install/transaction.c
@@ -228,6 +228,25 @@ out:
 	return 0;
 }
 
+static bool
+all_pkgs_on_hold(struct transaction *trans) {
+	xbps_object_t obj;
+	const char *action = NULL;
+	bool all_on_hold = true;
+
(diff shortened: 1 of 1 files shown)


bin/xbps-install/transaction.c
```diff
@@ -228,6 +228,28 @@ show_transaction_sizes(struct transaction *trans, int cols)
 	return 0;
 }
 
+static bool
+all_pkgs_on_hold(struct transaction *trans) {
+	xbps_object_t obj;
+	const char *action;
+	bool all_on_hold;
```
xtraeme: please `declare` and `initialize` in the same line, gets rid of extra useless lines.

xtraeme: `bool all_on_hold = true`

mbarbar:
Done. I wound up squashing the commits to clean them up.

PR #194, opened 2019-12-27
Do not silently update xbps on any install/update transaction.

### LINKED ISSUE #142: xbps-src first package installed after xbps update goes missing
If there is a bump to xbps itself, the next package xbps-src tries to install will now be the xbps update instead of the package it tried to install.

https://build.voidlinux.org/builders/x86_64-musl_builder/builds/21447/steps/shell_3/logs/stdio

Maybe we could add an EAGAIN exist status in this case so that xbps-src knows that it has to try to install the package again.

### LINKED ISSUE #166: xbps-install -Sy foo fails, performs unexpected action, exits zero
when i run `xbps-install -Sy foo`, i expect xbps to either install `foo` or fail.  instead, it does something else (updates libxbps/xbps) and claims success (exits zero) even though it clearly failed to do what was requested, not to mention it did something i didn't ask for *instead*.

the man page lacks any mention of this POLA violation, as well as recommendations regarding reliable unattended use.

xtraeme:
When there's a new xbps update, xbps-install(1) will now return
EBUSY (16) and a message (if dry-run disabled) explaining
how to proceed.

If there's an update and transaction does not contain xbps, it will
error out unless the 'xbps' pkg is the only target pkg, i.e:

```
	# xbps-install -u
        The 'xbps' package must be updated, please run `xbps-install -u xbps`
	# echo $?
	16
```

To update xbps, the only way to proceed is to explicitly declare
it as an update, i.e:

```
	# xbps-install -u xbps
```

The dry-run mode will still show there's an xbps update.

Modified the existing test cases to satisfy the new behaviour.

Closes #166
Closes #142

### DIFF
--- a/bin/xbps-install/transaction.c
+++ b/bin/xbps-install/transaction.c
@@ -1,5 +1,5 @@
 /*-
- * Copyright (c) 2009-2015 Juan Romero Pardines.
+ * Copyright (c) 2009-2019 Juan Romero Pardines.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -229,7 +229,8 @@ out:
 }
 
 static bool
-all_pkgs_on_hold(struct transaction *trans) {
+all_pkgs_on_hold(struct transaction *trans)
+{
 	xbps_object_t obj;
 	const char *action = NULL;
 	bool all_on_hold = true;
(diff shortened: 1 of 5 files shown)


xtraeme:
@Duncaen any objection?

PR #206, opened 2020-01-13
tests: add virtualpkg_multirepo

st3r4g:
I observed that `xbps-install` doesn't seem to honor a virtualpkg= override when the providers for the virtual are in multiple repositories. The package in the first specified repository wins regardless, so this may be a bug. This time I provided directly a test to reproduce the issue, I'm not sure if I put it in the right place.


xtraeme:
Thanks for the test case, as you can see in my commit the correct place was in vpkg_test.sh.

PR #209, opened 2020-01-18
Fix digest

Duncaen:
First commit fixes the checksum output if a download was continued.

Second commit adds a checksum output to `xbps-fetch(1)`, I mainly did this to easily test the fix but I think this could be useful for xbps-src because large downloads can be fetched and the checksum generated at the same time.


Duncaen:
@xtraeme what do you think, ca we use the checksum flag for `xbps-fetch` or should I drop the commit?

xtraeme:
sounds good to me as it is

PR #213, opened 2020-01-22
change to allow non-root user to use xbps-install --download-only --c…

dshmgh:
…ache-dir mycache

I added a check for the the root euid, since if the euid is root, the pkgdb could be written, so in that case we should still lock the pkgdb

Perhaps I need to look further at your code base as some of my edits go past 80 chars per line.


xtraeme:
Please fix the commit message, something like:

```short title: ...
...
```

xtraeme:
This logic can be simplified

PR #214, opened 2020-01-24
RFC: configuration: add keepconf option

andkem:
Add configuration option keepconf that stops xbps from overwriting
unchanged configuration files. If keepconf=true, xbps will store the new
configuration as <name>.new-<version> instead of overwriting unchanged
configuration files.

On production servers which I take care of, there are a number of different programs using default configuration. When upgrading, xbps normally overwrites an unchanged configuration with new defaults. This can lead to the server behaving unexpectedly. This is an RFC/PoC for adding a configuration knob that allows preventing xbps from overwriting default configuration, giving the user complete control.

Is this a good idea?

### DIFF
--- a/data/xbps.d.5
+++ b/data/xbps.d.5
@@ -1,4 +1,4 @@
-.Dd January 18, 2020
+.Dd February 05, 2020
 .Dt XBPS-D 5
 .Sh NAME
 .Nm xbps.d
@@ -84,6 +84,15 @@ Absolute path to a file and file globbing are supported, example:
 .It Sy preserve=/usr/bin/foo
 .It Sy preserve=/etc/foo/*.conf
 .El
+.It Sy keepconf=true|false
+If set to false (default), xbps will overwrite configuration files if
+they have not been changed since installation and a newer version is
+available.
+.Pp
+If set to true, xbps will save the new configuration file as
+<name>.new-<version> if the original configuration file has not been
+changed since installation.
+.Pp
 .It Sy repository=url
 Declares a package repository. The
 .Ar url
--- a/lib/package_config_files.c
+++ b/lib/package_config_files.c
@@ -170,11 +170,13 @@ xbps_entry_install_conf_file(struct xbps_handle *xhp,
 		/*
 		 * Orig = X, Curr = X, New = Y
 		 *
-		 * Install new file (installed file hasn't been modified).
+		 * Install new file (installed file hasn't been modified) if
+		 * configuration option keepconfig is NOT set.
 		 */
 		} else if ((strcmp(sha256_orig, sha256_cur) == 0) &&
 			   (strcmp(sha256_orig, sha256_new)) &&
-			   (strcmp(sha256_cur, sha256_new))) {
+			   (strcmp(sha256_cur, sha256_new)) &&
+			   (!(xhp->flags & XBPS_FLAG_KEEP_CONFIG))) {
 			xbps_set_cb_state(xhp, XBPS_STATE_CONFIG_FILE,
 			    0, pkgver,
 			    "Updating configuration file `%s' provided "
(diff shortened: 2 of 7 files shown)


Duncaen:
Sounds like a good idea.

Wondering if allowing to specify patterns using the new function the `skipextract` keyword introduced would make sense.

```
# keep all default configurations, execept /etc/foo.conf
keepconf=/*
keepconf=!/etc/foo.conf
```

lib/package_config_files.c
```diff
@@ -170,15 +170,30 @@ xbps_entry_install_conf_file(struct xbps_handle *xhp,
...
+			} else {
+				version = xbps_pkg_version(pkgver);
+				assert(version);
+				snprintf(buf, sizeof(buf), ".%s.new-%s", cffile,
+					version);
+				xbps_set_cb_state(xhp, XBPS_STATE_CONFIG_FILE,
+					0, pkgver, "Preserving `%s',  installing configuration file to `%s'.",
+					cffile, buf);
+				archive_entry_copy_pathname(entry, buf);
+			}
```
Vaelatern: I think I'd rather this logic be moved to only have the higher level `else if`, where you "fall through" to basically this exact code farther down only if the new configuration file isn't changed.

Vaelatern:
I assume everybody would agree this is a terrible default. The question is if it's good as a tunable. I think it is, but my comment above requesting that code be re-examined stands.

lib/package_config_files.c
andkem: I'm not sure I understand what you mean. Do you want to add a check for the configuration option not being set to enter this if-else and then always enter the last if-else (with the almost identical code) unconditionally if the option is set?

Something along the linens of the snippet below?
```suggestion
else if ((!(xhp->flags & XBPS_FLAG_KEEP_CONFIG)) &&
         (strcmp(sha256_orig, sha256_cur) == 0) &&
         (strcmp(sha256_orig, sha256_new)) &&
	 (strcmp(sha256_cur, sha256_new))) {`		
```

andkem:
@Duncaen Personally, I don't have any need of that kind of filtering, but if we can see a need, I wouldn't mind looking at adding the functionality.

lib/package_config_files.c
Vaelatern: Something like that yes

andkem: When considering how to write in the option, I did it the way it currently is because I liked the idea of the independent enumeration of configuration cases. I thought a minor code duplication was preferable over having the if-else-statements deal with more than one specific set of configuration clash as I don't like adding on "exceptions to the rules".

I'll see if I can write it in such a way that it doesn't get too convoluted, but I'm also open to the possibility that my preferences might be off.

Vaelatern: @Duncaen opinions?

andkem: I contemplated putting the preservation functionality in its own function at first, but there are a lot of variables being used that would lead to an enormous argument list.

Depending on how you feel about context specific macros, one solution could be simply putting the function in a macro that takes no arguments (except for maybe the log message if we care to differentiate between the two cases) and counts on the scope having correctly named variables. That way code duplication would be reduced, while the current control flow is kept.

xtraeme: Don't complicate it more than it is already, the suggested change above looks ok to me.

data/xbps.d.5
```diff
@@ -84,6 +84,15 @@ Absolute path to a file and file globbing are supported, example:
...
 .El
+.It Sy keepconf=true|false
+If set to false (default), xbps will overwrite configuration files if
+they have not been changed since installation and a newer version is
+available.
+.Pp
+If set to true, xbps will save the new configuration file as
+<name>.new-<version> if the original configuration file has not been
+changed since installation.
+.Pp
```
xtraeme: Please bump date in manpage

lib/package_config_files.c
andkem: Sure, done.

data/xbps.d.5
andkem: Fixed

xtraeme:
@Duncaen lgtm, any objection?

Duncaen:
looks god to me.

I'm not 100% on the naming and the descriptions.
I feel like `keepconf` as a keyword could be misinterpreted and I think simplifying the description to only explain the true case would make it easier to understand.

> The `keepconf` (disabled by default) keyword can be used to prevent
> xbps from overwriting configuration files.

Already gives a somewhat wrong description.

I think something simple would be better:

> If enabled unmodified configuration files will not be overwritten and instead extracted with a `.new-$version` suffix.

xtraeme:
Agreed

Duncaen:
Or maybe changing `true` and `false` to something like `always` and `modified` respectively, which would allow future additions like `never` and easily interpreted as "keep configuration files yes/no".

xtraeme:
I think this would complicate things. @Duncaen I don't have any issue with true/false, in fact it means what's expected.

andkem:
@Duncaen I wrote "[...] prevent xbps from overwriting configuration files." since I view the option from a black box perspective. In my opinion, the end user isn't really interested in the implementation details. I intended for the option to make xbps never overwrite a configuration file. That the only case that needed to be changed were the one where the default configuration had been unmodified, I see as an implementation detail.

Issue #215, opened 2020-01-25
Packages on hold don't get ignored while checking for shlibs


xtraeme:
I found this issue recently because I've got `sway` on hold due to the `elogind` build option. `wlroots` was updated to `0.10.0` (soname bump) and `sway` on hold.

The update unexpectedly went ok (apparently) but this allowed to have the sway pkg now broken due to missing symbols in `wlroots`.

A fix for this will appear in some minutes, with a proper test case.

PR #216, opened 2020-01-25
Fix for #215

xtraeme:
tests: rename downgrade_hold_test.sh to hold_test.sh.

Added new test case for https://github.com/void-linux/xbps/issues/215

This fails as expected right now because the fix is still
not there.


xtraeme:
And this is the real issue:

```
[DEBUG] Added `sway-1.4_1' into the dependency list (/tmp/kyua.EgblOP/178/work/repo)
[DEBUG] xbps_transaction_update_packages: trans_find_pkg sway-1.2_1: 0
```

PR #220, opened 2020-02-07
xbps-fetch: TIMEOUT_CONNECTION to customize time waiting for response

Chocimier:
Allows to [configure CI job](https://github.com/Chocimier/void-packages-org/commit/a09733b7748ede191a4cbf44bc447952d96bdf63) to build from ftp source without taking too much time as [demostrated with chrootkit](https://travis-ci.org/Chocimier/void-packages-org/builds/647493514).

### DIFF
--- a/bin/xbps-fetch/xbps-fetch.1
+++ b/bin/xbps-fetch/xbps-fetch.1
@@ -82,6 +82,10 @@ Overrides the default CA certificates path, by default set to
 Sets the SSL/TLS client verification certificate file.
 .It Sy SSL_CLIENT_KEY_FILE
 Sets the SSL/TLS client verification key file.
+.It Sy CONNECTION_TIMEOUT
+Sets connection timeout in milliseconds
+instead of default value of 5 minutes.
+When -1, waits indefinitely.
 .El
 .Sh SEE ALSO
 .Xr xbps-checkvers 1 ,
--- a/lib/fetch/common.c
+++ b/lib/fetch/common.c
@@ -428,6 +428,31 @@ fetch_socks5(conn_t *conn, struct url *url, struct url *socks, int verbose)
 	return 0;
 }
 
+static int
+get_conn_timeout(void)
+{
+	static int result = -2;
+	char *conn_timeout;
+
+	if (result != -2) {
+		return result;
+	}
+
+	conn_timeout = getenv("CONNECTION_TIMEOUT");
+	if (conn_timeout) {
+		char *char_read = conn_timeout;
+		long from_env = strtol(conn_timeout, &char_read, 10);
+		if (from_env < -1 || char_read == conn_timeout) {
+			from_env = fetchConnTimeout;
+		}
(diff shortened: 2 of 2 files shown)


xtraeme:
ok looks good at first glance

@duncaen ?

xtraeme:
```
=> chkrootkit-0.53_2: fetching distfile 'chkrootkit-0.53.tar.gz'...

ftp://ftp.pangeia.com.br/pub/seg/pac/chkrootkit-0.53.tar.gz: Can't open data connection

=> chkrootkit-0.53_2: fetch attempt 2 of 15...

chkrootkit-0.53.tar.gz: [40KB 10%] 33MB/s ETA: 00m00s

chkrootkit-0.53.tar.gz: 40KB [avg rate: 322MB/s]
```

FTR

bin/xbps-fetch/xbps-fetch.1
```diff
@@ -82,6 +82,9 @@ Overrides the default CA certificates path, by default set to
 Sets the SSL/TLS client verification certificate file.
 .It Sy SSL_CLIENT_KEY_FILE
 Sets the SSL/TLS client verification key file.
+.It Sy TIMEOUT_CONNECTION
+Sets connection timeout in milliseconds.
+When zero, default value is used.
```
xtraeme: what's default value? I don't know

Duncaen:
`CONNECTION_TIMEOUT` sounds more natural, this wouldn't allow 0 or -1 (infinite), I'll have to think about this tomorrow.

Chocimier:
Variable rename, -1 allowed, default timeout documented.

xtraeme:
I'm ok with this one, @Duncaen ?

xtraeme:
The default seems to be too high?

Duncaen:
I think that this is what curl used by default.

I guess the PR is fine, but I still think the xbps-src patch is wrong, depending on `xbps-fetch` behavior while everything else does not specifically depend on the fetch_cmd.

Chocimier:
> xbps-src patch is wrong, depending on xbps-fetch behavior

Didn't think of that, but it's not that bad. If fetch_cmd happens to be xbps-fetch, some settings are applied, in other cases nothing happens. Given that there is no standard way, other options I see are: enumerating other supported tools or ignoring problem at all.
Proposals how to improve are welcome, naturally.

xtraeme:
I still think that 5 minutes is too much. How about changing it to 1 minute?

xtraeme:
Also while it's ok to have the retry code in xbps-src, I would like to have this in `xbps-fetch(1)` instead. This is already in `TODO`.

xtraeme:
unless this gets any progress it will have to wait until 0.60, because I plan to release 0.59 tomorrow.

Chocimier:
Using what curl uses sound ok, but I can change if you insist. But not the day before release.

xtraeme:
If this is what curl uses OK, fine with me.

PR #223, opened 2020-02-10
libxbps: avoid malloc in xbps_file_hash{,_raw}.

xtraeme:
The xbps_file_hash() and xbps_file_hash_raw() functions
have been modified to not return a heap allocation;
instead the caller must supply a buffer with enough size.

These functions now use the following prototype:

bool xbps_file_hash(char *dst, size_t len, const char *file)
bool xbps_file_hash_raw(unsigned char *dst, size_t len, const char *file)

cc @duncaen

### DIFF
--- a/bin/xbps-create/main.c
+++ b/bin/xbps-create/main.c
@@ -437,6 +437,7 @@ ftw_cb(const char *fpath, const struct stat *sb, const struct dirent *dir UNUSED
 		xbps_object_t obj;
 		xbps_dictionary_t linkinfo;
 		uint64_t inode = 0;
+		char digest[128];
 		/*
 		 * Regular files. First find out if it's a hardlink:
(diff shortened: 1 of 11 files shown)


xtraeme:
While we are at it, I propose to rename these funcs to `xbps_file_hash_sha256{,_raw}`.

xtraeme:
hmm wait, there are some things that are broken now.

Duncaen:
lol I did the same thing. http://ix.io/2biy

Duncaen:
I think renaming is a good idea, if we ever want to switch to i.e. blake2 we can't really just change the `xbps_*_hash_*` functions.

xtraeme:
omg, I didn't know you were working on this.

xtraeme:
I prefer your diff, please submit a PR. Closing this

PR #226, opened 2020-02-12
xbps_get_pkg_fulldeptree: switch to uthash.

xtraeme:
There are still two users that can benefit from
uthash: transaction_files.c and xbps-fbulk.

cc @duncaen

### DIFF
--- a/3RDPARTY
+++ b/3RDPARTY
@@ -19,3 +19,6 @@ internal use in the code:
 
 - portableproplib-0.6.8 (lib/portableproplib) from
   https://github.com/xtraeme/portableproplib
+
+- uthash-2.1.0 (include/uthash.h) from
+  http://troydhanson.github.com/uthash/
--- a/include/uthash.h
+++ b/include/uthash.h
@@ -1,0 +1,1150 @@
+/*
+Copyright (c) 2003-2018, Troy D. Hanson     http://troydhanson.github.com/uthash/
+All rights reserved.
+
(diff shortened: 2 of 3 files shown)


Duncaen:
Looks good.

* Maybe we can change the api so malloc/calloc errors could be handled?
* Maybe add a internal structure that holds the hash head and the results array so that if we ever want to get full dep trees multi threaded we don't run into issues, would this make sense?

xtraeme:
@Duncaen 

- uthash just calls exit(-1) if malloc fails
- not sure if we ever run fulldeptree multithreaded

Duncaen:
Ok.

>     * uthash just calls exit(-1) if malloc fails

Hm I think we might have to add our own handler if we use uthash somewhere inside the transaction to at least try to recover from it, for now this looks ok.

xtraeme:
@Duncaen I agree, I thought about this. If in the future we need this, we could set uthash hooks for malloc/free.

xtraeme:
@Duncaen been thinking about this... we could get rid of the transaction dictionary and use a hash table with something like:

```
struct pkg {
   xbps_dictionary_t d;
   const char *pkgver;
   UT_hash_handle hh;
}
```

Duncaen:
Yes I would like to do that, I've been playing around with that a bit.

[types.txt](https://github.com/void-linux/xbps/files/4192864/types.txt)

Issue #187, opened 2019-12-09
Overall progress bar or percentage when downloading


mbarbar:
A progress bar showing how much has been downloaded overall would be nice. Currently, when I want to know how much has been downloaded, I kill xbps and start it again.

I'm happy to implement it. If there is interest, I have two concerns:

1. It seems to me that the total downloaded data, across packages, is not kept anywhere. Assuming I just couldn't find that info, it seems the most appropriate (global) place to store this would be in the `transd` dictionary in `xbps_handle`. Is there a better place?

2. I see two possibilities with respect to UI.

* At the beginning of the download lines, for example,
```
[  7%]pkg1.x86_64.xbps: [723KB 0%] 25MB/s ETA: 00m00s
...
...
[ 54%]pkg30.x86_64.xbps: [723KB 0%] 25MB/s ETA: 00m00s
```

* Underneath the download/verifying signature line. It would then always be the final line of the output. For example,

 ```
pkg2.x86_64.xbps: [723KB 0%] 25MB/s ETA: 00m00s
[  9% ***---------------------------]
...
...
pkg25.x86_64.xbps: [723KB 0%] 25MB/s ETA: 00m00s
[ 34% **********--------------------]
```

xtraeme:
Sounds good, do you have a patch? :+1:

xtraeme:
I prefer the second option (underneath)

mbarbar:
Caught up with work till ~mid January, hopefully a little earlier, then I'll get into it.

sineemore:
> It seems to me that the total downloaded data, across packages, is not kept anywhere.

@mbarbar transaction dictionary has "total-download-size" key.
@xtraeme is it the value we should use?

sineemore:
Seems like on possible problem with storing this in transaction dictionary, is that it is immutable.

xtraeme:
`total-download-size` is the right obj

xtraeme:
https://github.com/void-linux/xbps/blob/master/lib/transaction_prepare.c#L92

xtraeme:
I see what's the problem, `dlsize` is only incremented with `xhp->flags & XBPS_FLAG_DOWNLOAD_ONLY`. I'm adding a fix

xtraeme:
The transaction dictionary now contains proper info, including `total-download-size`.

xtraeme:
#230 implements the first suggested option shown in this PR.

And it seems to be good enough.

xtraeme:
[...]

PR #230, opened 2020-02-16
xbps-install: show overall download progress

sineemore:


sineemore:
Was proposed in #187

sineemore:
Implementation seems a bit quirky in a way it stores downloaded size.

bin/xbps-install/fetch_cb.c
```diff
@@ -145,9 +168,9 @@ stat_display(const struct xbps_fetch_cb_data *xfpd, void *cbdata)
 		(void)xbps_humanize_number(totsize, (int64_t)xfpd->file_size);
 	}
 	if (v_tty)
-		fprintf(stderr, "%s: [%s %d%%] %s ETA: %s\033[K\r",
-		    xfpd->file_name, totsize, percentage,
-		    stat_bps(xfpd, xfer), stat_eta(xfpd, xfer));
+		fprintf(stderr, "%s %s: [%s %d%%] %s ETA: %s\033[K\r",
```
sineemore: Progress is embedded within download line. I've tried adding it as a separate line, but it messes up by package verify message and I don't see a good (a simple) way to address this.

xtraeme:
I just tested this, it works as expected with my last two changes to master.

bin/xbps-install/fetch_cb.c
xtraeme: I think it's good enough

xtraeme:
@Duncaen ?

Issue #35, opened 2018-12-09
zstd compression by default propose


whoizit:
https://bugs.archlinux.org/task/59081
```
757M pycharm-professional-2018.1.4-1-x86_64.pkg.tar
338M pycharm-professional-2018.1.4-1-x86_64.pkg.tar.zst
309M pycharm-professional-2018.1.4-1-x86_64.pkg.tar.xz
```
```
# compress
$ time zstd -T0 pycharm-professional-2018.1.4-1-x86_64.pkg.tar
real 0m2,941s
$ time xz -T0 -1 pycharm-professional-2018.1.4-1-x86_64.pkg.tar
real 0m31,260s
```
```
# decompress
$ time zstd -d pycharm-professional-2018.1.4-1-x86_64.pkg.tar.zst
real 0m1,869s
$ time xz -d pycharm-professional-2018.1.4-1-x86_64.pkg.tar.xz
real 0m18,694s
```

travankor:
https://github.com/vasi/pixz would have faster decompression than zstd on most modern processors (and have a higher compression ratio).

https://github.com/facebook/zstd/tree/dev/contrib/pzstd with +19 compression would make more sense IMO

whoizit:
@travankor can u post any link about zstd 2.4.X? can't found it

bobertlo:
@whoizit I think he meant 1.4.X. The latest release mentions stabilizing the API in the next major release (1.4.X) https://github.com/facebook/zstd/releases/tag/v1.3.8 

I would guess libarchive support would be very helpful to getting this merged into xbps. The last release was two years ago, and the zstd support has been in master since right after that release.

whoizit:
but zst already support in libarchive 3.3.3

bobertlo:
Oh, sorry. The latest two release tags were squashed above older releases on github. I can't speak for the infrastructure team, but the 10% increase in storage and bandwidth is significant. That said, the speed increase and anecdotal evidence from @travankor is quite compelling also. Compression speed is not exactly a major cost in the build-chain, but the speed-up for end users could be impressive.

xtraeme:
Well, I think it should be way faster... but xz provides higher compression ratio.

xtraeme:
I'm on it

xtraeme:
I implemented this, and as bonus repodata is now zstd compressed by default.

See commits:
- https://github.com/void-linux/xbps/commit/ab9a587d34e945b7920f363403d90565c1f79502
[...]

Issue #234, opened 2020-02-21
xbps-remove -R removes manually installed deps


travankor:
Not sure if this reproducible for other people, but:
```
doas xbps-install ffmpeg
doas xbps-install geeqie
doas xbps-remove -R geeqie
```
results in xbps wanting to remove ffmpeg, which was manually installed.

After installing `geeqie`, `xbps-query -m` still has `ffmpeg` in the list. Also, if I do `xbps-remove geeqie` with no flags, `ffmpeg` is not considered an orphan.

xtraeme:
please try with up-to-date master

xtraeme:
also, you can use `-n, --dry-run` that doesn't need root perms and will give us the expected result

xtraeme:
I want `xbps-foo -V` output, of the affected command, and debug output with `-d`

travankor:
Current xbps is whatever is latest in the repos (not master):
```
xbps-remove -V
XBPS: 0.58 API: 20200118 GIT: UNSET
```

travankor:
`xbps-remove -dnR geeqie`
http://ix.io/2cii

xtraeme:
This is not good, you should have the same result than I do:

```
[juan@leysa void-packages]$ xbps-query -V
XBPS: 0.59 API: 20200221 GIT: a16e695d
```

See https://github.com/void-linux/xbps#building-and-testing-for-dummies

travankor:
Bug still seems present on master.
```
xbps-query -V
XBPS: 0.59 API: 20200221 GIT: a16e695
```

```
xbps-remove -dnR geeqie
http://ix.io/2cin
```

Duncaen:
`xbps_dictionary_get_bool` (`prop_dictionary_get_bool`) doesn't set the boolean if the object in the dictionary is not a boolean (or doesn't exist).
The code doesn't check the return value of it and therefor `automatic` will be always be true when it finds a one package that is automatically installed in the revdeps and then considers all following packages in the revdeps automatically installed.

https://github.com/void-linux/xbps/blob/master/lib/package_orphans.c#L181
```
	for (unsigned int i = 0; i < xbps_array_count(array); i++) {
		︙
		bool automatic = false;
		︙
		xbps_dbg_printf(xhp, " processing rdeps for %s\n", pkgver);
		for (unsigned int x = 0; x < xbps_array_count(rdeps); x++) {
			︙
[...]

PR #235, opened 2020-02-21
xbps-{create,rindex}: use zstd by default.

xtraeme:
There's lots of benchmarks right there, zlib vs zstd, xz vs zstd.
Take your own conclusion.

### DIFF
--- a/bin/xbps-create/main.c
+++ b/bin/xbps-create/main.c
@@ -105,7 +105,7 @@ usage(void)
 	"                     This expects a blank separated list of <name>:<symlink>:<target>, e.g\n"
 	"                     'vi:/usr/bin/vi:/usr/bin/vim foo:/usr/bin/foo:/usr/bin/blah'.\n"
 	" --build-options     A string with the used build options.\n"
-	" --compression       Compression format: none, gzip, bzip2, lz4, zstd, xz (default).\n"
+	" --compression       Compression format: none, gzip, bzip2, lz4, xz, zstd (default).\n"
 	" --shlib-provides    List of provided shared libraries (blank separated list,\n"
 	"                     e.g 'libfoo.so.1 libblah.so.2').\n"
 	" --shlib-requires    List of required shared libraries (blank separated list,\n"
@@ -1050,10 +1050,14 @@ main(int argc, char **argv)
 	if (ar == NULL)
 		die("cannot create new archive");
 	/*
-	 * Set compression format, xz if unset.
+	 * Set compression format, zstd by default.
 	 */
-	if (compression == NULL || strcmp(compression, "xz") == 0) {
+	if (compression == NULL || strcmp(compression, "zstd") == 0) {
+		archive_write_add_filter_zstd(ar);
+		archive_write_set_options(ar, "compression-level=9");
+	} else if (strcmp(compression, "xz") == 0) {
 		archive_write_add_filter_xz(ar);
+		archive_write_set_options(ar, "compression-level=9");
 	} else if (strcmp(compression, "gzip") == 0) {
 		archive_write_add_filter_gzip(ar);
 		archive_write_set_options(ar, "compression-level=9");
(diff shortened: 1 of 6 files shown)


xtraeme:
Closes #35

bin/xbps-create/main.c
```diff
@@ -1050,10 +1050,14 @@ main(int argc, char **argv)
 	if (ar == NULL)
 		die("cannot create new archive");
 	/*
-	 * Set compression format, xz if unset.
+	 * Set compression format, zstd by default.
 	 */
-	if (compression == NULL || strcmp(compression, "xz") == 0) {
+	if (compression == NULL || strcmp(compression, "zstd") == 0) {
 		archive_write_add_filter_xz(ar);
```
Duncaen: This is still `_xz`.

Duncaen:
Not sure how I feel about this, how long has zstd been supported? What would be the timeframe in which a users who didn't update yet would have to do a update using static binaries?

bin/xbps-create/main.c
xtraeme: omg you are right, I forgot to change this, thanks

Chocimier:
We could stay compatible by still using xz for index, xbps and all its dependencies - with compression per package option analogical to current build options per package.
Time of decompressing index is less than time of typing `xi -Su`. I do not think decreasing it is worth breaking updates.

bin/xbps-create/main.c
```diff
@@ -1050,10 +1050,14 @@ main(int argc, char **argv)
...
+		archive_write_set_options(ar, "compression-level=19");
+	} else if (strcmp(compression, "xz") == 0) {
+		archive_write_add_filter_xz(ar);
+		archive_write_set_options(ar, "compression-level=19");
```
Chocimier: Shouldn't be 9 for xz?

xtraeme:
@Duncaen xbps supports reading zstd archives since 0.54.

xtraeme:
so we can expect all users to have this version at least. Live images starting since 2019 should work with no issues.

Duncaen:
This is only the latest ISO then, 0.54 was released 20190620.

I think I like the idea of having a `xbps-src` option to change the compressions for xbps and the packages xbps depends and keeping repodata as gzip for a few more 

I guess 6 months for rolling release is enough, everything older requiring manual work with static binaries isn't too bad.

xtraeme:
`xbps-src` already supports this for a long time, I use:

```
[juan@leysa xbps]$ grep -i comp ~/.xbps-src.conf
XBPS_REPO_COMPTYPE=zstd
XBPS_PKG_COMPTYPE=zstd
[juan@leysa xbps]$
```

Duncaen:
I mean use this on a per package basis if we would decide to not break compatibility with > 6 month outdated systems.

xtraeme:
ah, sure.

xtraeme:
hmm I think I'll add support to set compression level via environment. This way one could do:

```
$ COMPLEVEL=22 xbps-create ...
```

or something like this

xtraeme:
Hmm, there's no >=0.54 available at https://a-hel-fi.m.voidlinux.org/static/

xtraeme:
Can anyone please push 0.58?

xtraeme:
This way we can make always a switch on old systems

Duncaen:
There is 0.56 and 0.57, but I can make new ones.

xtraeme:
oh true. Yes, please do. 0.58 is good enough

xtraeme:
```
xbps-static-latest.aarch64-musl.tar.xz             12-Dec-2015 07:45             1679040
xbps-static-latest.armv6l-musl.tar.xz              12-Dec-2015 07:45             1613140
xbps-static-latest.i686-musl.tar.xz                12-Dec-2015 07:45             2281348
xbps-static-latest.mips-musl.tar.xz                12-Dec-2015 07:45             2324644
xbps-static-latest.mipsel-musl.tar.xz              12-Dec-2015 07:45             2454672
xbps-static-latest.x86_64-musl.tar.xz              12-Dec-2015 07:45             2105416
```

please point latest to latest :-)

Duncaen:
I do think https://github.com/void-linux/xbps/issues/234 should be fixed with with a patch in void-packages, before we make a new release.

Duncaen:
Updated the static builds for the architectures we have, i686 is missing because we don't build the musl package.

xtraeme:
@Duncaen thanks!

PR #237, opened 2020-02-21
xbps_find_pkg_orphans: fix for #234

### LINKED ISSUE #234: xbps-remove -R removes manually installed deps

xtraeme:
Make sure "automatic" bool is initialized to false before
checking its value. This way if xbps_dictionary_get_bool()
fails, "automatic" will be set to false.

Closes #234


xtraeme:
hmm we could have a test case for this one too.

PR #240, opened 2020-02-22
Use uthash instead of our own hashtables

Duncaen:


xtraeme:
could you also please adapt xbps-fbulk?

Duncaen:
Maybe I do this in a separate PR, I think I have a few ideas to make xbps-fbulk suitable as buildbot replacement.

Duncaen:
I switched it to uthash, I think its better to keep `xbps-fbulk` as it is and maybe introduce a new tool.

xtraeme:
@Duncaen great! you rock :+1:

xtraeme:
@Duncaen the following diff fixes the -Wcast-qual error in uthash with clang:

```
[juan@leysa xbps]$ git diff
diff --git a/include/uthash.h b/include/uthash.h
index 76bdca64..06e4901d 100644
--- a/include/uthash.h
+++ b/include/uthash.h
@@ -144,7 +144,7 @@ typedef unsigned char uint8_t;
 /* calculate the element whose hash handle address is hhp */
 #define ELMT_FROM_HH(tbl,hhp) ((void*)(((char*)(hhp)) - ((tbl)->hho)))
 /* calculate the hash handle from element address elp */
-#define HH_FROM_ELMT(tbl,elp) ((UT_hash_handle *)(((char*)(elp)) + ((tbl)->hho)))
+#define HH_FROM_ELMT(tbl,elp) ((UT_hash_handle *)(void *)(((char*)(elp)) + ((tbl)->hho)))

 #define HASH_ROLLBACK_BKT(hh, head, itemptrhh)                                   \
 do {                                                                             \
[juan@leysa xbps]$
```

PR #255, opened 2020-03-31
zsh-completion: update archs, add 'sort-dependencies'

Piraty:


xtraeme:
@Piraty I just fixed the travis issue. We were missing `stat(1)`.

xtraeme:
hmm wait, still fails hmm

xtraeme:
https://travis-ci.org/github/void-linux/xbps/jobs/669154357#L1514

This is the real issue:
```
stat: cannot statx 'root/file': Operation not permitted
```

xtraeme:
hmm maybe something has changed in the docker image? dunno what's the issue, but we didn't change anything relevant for the previous commits.

@Duncaen any idea?

xtraeme:
and it seems to only happen with glibc

xtraeme:
hmm I think there was a coreutils update in that time

xtraeme:
ok testing a fix

xtraeme:
definitely it's a change in how the new coreutils handles extended attributtes.

xtraeme:
from coreutils-8.32:

```
  stat(1) supports a new --cached= option, used on systems with statx(2)
  to control cache coherency of file system attributes,
  useful on network file systems.

** Improvements

  stat and ls now use the statx() system call where available, which can
  operate more efficiently by only retrieving requested attributes.
```

xtraeme:
coreutils code mentions that `statx` needs `glibc>=2.28` and `linux>=4.11`.

I think travis host does not meet this requirements and statx always fails with EACCES/EPERM.

I'm disabling statx support in coreutils with glibc for now. Or any other idea?

PR #257, opened 2020-03-31
tests/configure: make rootdir before setting umask.

xtraeme:
Let's see if the issue is umask or the underlaying fs in travis.


xtraeme:
@Duncaen see how with glibc `statx` always fails, even without `umask`.

xtraeme:
`stat --cached` does not have any effect

Issue #259, opened 2020-04-03
xbps-rindex -a: validate deps


xtraeme:
I'm working in a feature that I had in my mind for a while.

`xbps-rindex -a` to add a package to the repo index, does not validate dependencies, thus a package with unresolved dependencies can be added and nobody would notice until it's installed.

This feature will modify `xbps-rindex -a` to always validate its required dependencies before the pkg is registered into the repository index.

st3r4g:
How would it work when the dependency is in another repo (e.g. `-dbg` packages)?

xtraeme:
via rpool

xtraeme:
the same strategy will have to be made for `xbps-rindex -c`. While cleaning up the repo index, before unindexing an entry it will have to verify its revdeps.

xtraeme:
this is the theory, let's see if I can come up with something that works :-)

xtraeme:
another interesting feature would be adding a new mode that does the validation for existing repos.

xtraeme:
ok found an issue in the first implementation, it just needs to verify all pkgs at once before flushing repodata :-) otherwise if you do `xbps-rindex -a repo/foo.xbps` its dependencies might not be there...

xtraeme:
thinking about it, this should be enabled via a new flag. Because while building everything from scratch the validation will be too strict.

xtraeme:
hmm this is not that easy, but I still think it's doable

Duncaen:
I think this would be a nice opportunity to introduce a separate tool.

* `xbps-rindex` is the tool for `xbps-src` and local repositories (always staged).
* `xbps-repoadmin` is responsible for everything related to publishing repositories.

## `xbps-rindex`

* adds packages to the architectures `-stagedata`
* remove signing staging etc.

## `xbps-repoadmin`/`xbps-repodb`?

* works on multiple `-stagedata` files.
* checks shlibs and dependencies and then selectively moves satisfied packages into repodata.
* this tool is also responsible for signing it will always be executed with the signing keys available.
* if it moves a package from `-stagedata` to `-repodata` it signs them.
* in the future this tool will also be able to sign the `-repodata` files whenever it changes them.
* this tool also does the cleanup stuff, it knows multiple repositories, knows `-repodata` and `-stagedata` and can do a better job at cleaning them up.

[...]

PR #260, opened 2020-04-03
xbps-rindex: multiple changes to validate dependencies.

xtraeme:
This patchset makes xbps-rindex(1) validate pkg dependencies
of all packages stored in a repository, taking into account
all repositories registered (rpool) or in current repo (if rpool
does not contain any registered repo).

This now contains some changes in its supported options:

 - Added -C/--confdir to set xbps.d configuration directory.
 - Added --validate long option to enable this behaviour with -a/--add.
 - Previous -C short option has been renamed to -H, --hashcheck.

Added two test cases to verify both behaviours: single and multirepo.

For https://github.com/void-linux/xbps/issues/259

### DIFF
--- a/bin/xbps-rindex/defs.h
+++ b/bin/xbps-rindex/defs.h
@@ -67,7 +67,7 @@
 #define _XBPS_RINDEX		"xbps-rindex"
 
 /* From index-add.c */
-int	index_add(struct xbps_handle *, int, int, char **, bool, const char *);
+int	index_add(struct xbps_handle *, int, int, char **, bool, bool, const char *);
 
 /* From index-clean.c */
 int	index_clean(struct xbps_handle *, const char *, bool, const char *);
--- a/bin/xbps-rindex/index-add.c
+++ b/bin/xbps-rindex/index-add.c
@@ -1,5 +1,7 @@
 /*-
- * Copyright (c) 2012-2015 Juan Romero Pardines.
+ * Copyright (c) 2012-2020 Juan Romero Pardines.
+ * Copyright (c) 2016 Enno Boland <gottox@voidlinux.org>
+ * Copyright (c) 2016-2020 Duncan Overbruck <mail@duncano.de>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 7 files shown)


Duncaen:
I see the value in this, but because it can only commit to one repository I fail to see how this can effectively be used with dependencies over multiple repos.
This would work with most packages, but with multilib, debug and nonfree I don't really see how we could make use of it in xbps-src.

Thats why I proposed a separate tool which can commit chunks of stagedata to each repodata.

xtraeme:
@Duncaen it will just work with xbps-src, due to xbps.d(5). All repos are in rpool while xbps-rindex is executed in the chroot.

Duncaen:
How can this work, with debug you have to add packages to multiple repositories.

One of the following will fail or needs to be run without `--validate`:

* `xbps-rindex --validate -a debug./*-dbg.*.xbps` 
* `xbps-rindex --validate -a *.xbps`.

xtraeme:
@Duncaen a separate tool will come later.  Definitely we want this feature for the new tool, as well as shlib checking.

xtraeme:
@Duncaen see the multirepo test cases.

xtraeme:
@Duncaen the main repo will always have to be processed first, and then those repos that require it will resolve deps.

xtraeme:
main -> debug
main -> nonfree
main -> multilib -> nonfree -> whatever

xtraeme:
it's very important that main repo does not depend on any other repo!

Duncaen:
So the following cases would break?

* main package depeneds on -32bit packages (gcc-multilib)
* whenever there is a big rebuild ongoing, do you hold packages back and only register them later when all are built? If so, how would xbps-src depend on incomplete builds or know if a package has already been build but not registered?

Duncaen:
Ok 2 is probably not a problem because it only validates packages that are about to be added, not if something breaks already added packages.

xtraeme:
@Duncaen that's why it's not enabled by default, in `xbps-src bootstrap` there might be pkgs with incomplete deps at some point.

The `gcc-multilib` pkg is a special case, and yes needs to be fixed, maybe by not enabling `--validate` at all!

xtraeme:
@Duncaen a solution would be moving `gcc-multilib` to the `multilib` repo.

Duncaen:
IMHO its now worth it to change xbps-rindex if we would remove the features again anyways, because I don't think this can be solved with just xbps-rindex.

xtraeme:
@Duncaen I'll start repoadmin in a short while :-)

xtraeme:
For reference, https://github.com/void-linux/xbps/issues/259#issuecomment-608367884

This is what I'll work on this evening.

PR #270, opened 2020-04-18
bin/xbps-query: simplify search result printing (with truncation)

Duncaen:
This should also fix a useless error from lgtm

### DIFF
--- a/bin/xbps-query/search.c
+++ b/bin/xbps-query/search.c
@@ -49,47 +49,42 @@ struct search_data {
 	unsigned int maxcols;
 	const char *pat, *prop, *repourl;
 	xbps_array_t results;
+	char *linebuf;
 };
 
 static void
 print_results(struct xbps_handle *xhp, struct search_data *sd)
 {
-	const char *pkgver = NULL, *desc = NULL, *inststr = NULL;
-	char tmp[256], *out;
-	unsigned int j, tlen = 0, len = 0;
+	const char *pkgver = NULL, *desc = NULL;
+	unsigned int align = 0, len;
 
 	/* Iterate over results array and find out largest pkgver string */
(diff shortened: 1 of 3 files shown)


xtraeme:
cool! what's going with travis?

bin/xbps-query/search.c
```diff
@@ -49,47 +49,34 @@ struct search_data {
...
-			printf("%s %s %s\n", inststr, tmp, desc);
+		len = snprintf(sd->linebuf, sd->maxcols, "[%s] %-*s %s",
+		    xbps_pkgdb_get_pkg(xhp, pkgver) ? "*" : "-",
+		    align, pkgver, desc);
+		/* add ellipsis if the line was truncated */
+		if (len >= sd->maxcols && sd->maxcols > 4) {
+			for (unsigned int j = 0; j < 3; j++)
+				sd->linebuf[sd->maxcols-j-1] = '.';
 		}
+		puts(sd->linebuf);
```
xtraeme, line 79: rofl this reminded me about tcl puts().

Duncaen:
restarted the travis builds, no idea what happened.

Duncaen:
Failed again for some reason oO.

we could also use unicode ellipsis (…) so we would only have to truncate one character, but that would require to allocate again maxcols+3 to have enough bytes for the unicode character at the end.

xtraeme:
do we have to care for 3 or 4 bytes? just allocate string*2 and be done rofl

Duncaen:
The calculation would again trigger the lgtm warning lol. Probably not worth getting 2 characters more from truncated output anyways.

xtraeme:
We just really need to know string length for output less 5 chars; if result-5 > maxcols; then result = result + "...\n"

PR #273, opened 2020-04-22
xbps_register_pkg: keep stored props in pkgdb on updates.

xtraeme:
I noticed that while updating a pkg that is on hold
or in repolock mode, does not keep those properties.

Always set those props in the new pkg dictionary to respect
this behaviour. If there's a pkg on hold and you update it,
you want to keep it in this state unless you tell it to change.

Added new test case to verify.

### DIFF
--- a/lib/package_register.c
+++ b/lib/package_register.c
@@ -1,5 +1,5 @@
 /*-
- * Copyright (c) 2008-2014 Juan Romero Pardines.
+ * Copyright (c) 2008-2020 Juan Romero Pardines.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -35,7 +35,8 @@ int HIDDEN
 xbps_register_pkg(struct xbps_handle *xhp, xbps_dictionary_t pkgrd)
 {
 	xbps_array_t replaces;
-	xbps_dictionary_t pkgd;
+	xbps_dictionary_t pkgd, pkgdbd;
+	xbps_object_t obj;
 	time_t t;
 	struct tm tm;
 	struct tm *tmp;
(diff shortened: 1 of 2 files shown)


Duncaen:
hm should explicitly updating packages on hold even work?
I feel like this kinda makes sense to support but on the other hand its a bit unexpected when scripting xbps installations.

xtraeme:
@Duncaen can you explain?

See the test case on this PR

xtraeme:
I think the following is a bug (this is what this PR fixes):

```
$ xbps-install foo
$ xbps-pkgdb -m hold foo
$ xbps-install -u foo
$ xbps-query -p hold foo
[prop lost]
```

xtraeme:
The same happened for more props: `repolock` and `automatic-install`.

Duncaen:
I just think that if hold is set it means hold and even manually trying to update the package should not update the package.

For repolock this seems correct as long as it updates from the locked repository.

xtraeme:
@Duncaen how would you update the pkg then?

xtraeme:
Ah I see what you mean. This would break the behaviour has been used for a long while.

I don't have any issue in changing it. Somehow I agree that if pkg is on hold, it should not be updated by any means, but this is what we've been using until now.

Duncaen:
Hm maybe allow it with `-f`, or has this already other implications with `-u`.

xtraeme:
yes, `-f` would overwrite unmodified files as well on install/update.

Duncaen:
With one `-f` it looks like it just makes the `reinstall` argument for `xbps_transaction_install_pkg` true. It only adds `XBPS_FLAG_FORCE_UNPACK` if the flag is specified twice.

xtraeme:
hmm true

xtraeme:
Maybe we should change the behaviour to:

```
$ xbps-pkgdb -m hold foo
$ xbps-install -u [foo]
[no update]
$ xbps-install -uf foo
[update]
```
This would mean explicitly update even if it's on hold. Do you agree?

xtraeme:
Even so, this PR is still valid because in any case the props must be kept on updates.

Duncaen:
Yes I think this would make more sense and works well with how force already works for installing packages.

> Even so, this PR is still valid because in any case the props must be kept on updates.

Yes PR is good, this is a separate issue.

Duncaen:
tbh I don't get why `xbps-install -u pkg` is not the same as `xbps-install -u` where a package is updated implicit, why is there a difference?

xtraeme:
ok, will submit a PR for the other behaviour after this one.

xtraeme:
I don't remember, maybe it was a workaround for hold.

xtraeme:
Pushing this one, cheers

Duncaen:
Looking at this, I think the bug needs to be fixed differently, a lot earlier otherwise it seems like repolock wouldn't work/

xtraeme:
We've got test cases for repolock already, it should work?

xtraeme:
Hmmm found a simpler way to fix this! omg

Duncaen:
hm yea just wrote a testcase for repolock with `-u A` this seems to work fine, I'll just push it to have another test case.

PR #274, opened 2020-04-22
Perform action with pkgs on hold only if -f is set.

xtraeme:
As discussed with @duncaen in #273 we've decided it's better
to only update packages on hold mode when `-f` is set.

That means that `xbps-install -u foo` won't work anymore
with pkgs on hold unless `-f` is set.

This way one could perform a full system upgrade even with
pkgs on hold via `xbps-install -fu`.

cc @duncaen

### DIFF
--- a/bin/xbps-install/main.c
+++ b/bin/xbps-install/main.c
@@ -155,6 +155,7 @@ main(int argc, char **argv)
 			break;
 		case 'f':
 			fflag++;
+			flags |= XBPS_FLAG_FORCE_INSTALL;
 			if (fflag > 1)
 				flags |= XBPS_FLAG_FORCE_UNPACK;
 			reinstall = true;
--- a/include/xbps.h.in
+++ b/include/xbps.h.in
@@ -51,7 +51,7 @@
  *
  * This header documents the full API for the XBPS Library.
  */
-#define XBPS_API_VERSION	"20200414"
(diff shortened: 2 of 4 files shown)


Duncaen:
I think I would change the `reinstall` in  `trans_find_pkg` and `xbps_transaction_install_pkg` parameter to `force` and add a new force parameter to `xbps_transaction_update_pkg`.

Not sure if `-fu` is a good idea, I've seen people do this for some stupid reason.

Duncaen:
This makes the api more selective and doesn't force update hold revdeps.

Duncaen:
I guess if you think `-fu` for distupgrade is usefull then maybe add a force parameter to `xbps_transaction_update_packages` too.

xtraeme:
I did it this way to avoid breaking ABI, but ok.

Duncaen:
We've been breaking ABI a lot lately so I'm not really concerned, I just think that it its a lot better to be able to selective force in this case, `xbps-install -f foo` behaves like that already so I think its better to match this.
This may also be a better solution for the future and from a API standpoint in case anyone actually uses the API directly.

xtraeme:
I also wanted to get rid of this reinstall arg anyway :-)

xtraeme:
Hmm maybe it's better to forbid `-fu`. Lots of users wouldn't expect this to update pkgs on hold...

PR #277, opened 2020-04-29
[WIP] Add orderby flag.

### LINKED ISSUE #163: [Feature request]: Sort manually installed packages by date.
Hi. with `-m` option in `xbps-query`, it just shows manually installed packages alphabetically. It can be sometimes useful to sort __manually__ installed packages by date.

### LINKED ISSUE #241: Another column in xbps-query result?
First, thanks for your nice distro and pkg manager!

Right now I use "xbps-query -Rs -" to get a list of all available packages for OctoXBPS.
Could it be possible to add "Download size" column in the result list?

Thanks and keep up the good work!

ericonr:
Initial implementation, only works in conjunction with the `-m` switch,
for now.

Will fix #163 

This is just a very rough start, but it had a sorted output :)

A few TODO comments around, too. And of_free isn't run if the key doesn't exist, it just exits.

I wasn't very sure of how proplib stuff worked, so I ended up duplicating all the strings.

Unless I implement an orderby function for each kind of listing, we'd need to make some changes to listing functions. I was thinking of adding a "mode" parameter in a parameter struct, perhaps. Something like:

- Unbuffered mode: the current one, just prints the packages to stdout.
- Formatted mode: to fix #241 (could be implemented together with unbuffered)
- Sorting/post processing mode: stores some strings or the dictionary obj themselves, then sorts/does something with them afterwards, and then outputs them using the format functions.

Taking all feedback, my hosted C needs improvement :)


ericonr:
@Duncaen I have to do some revision and I'd like to write some tests for it first, but I think it's in good shape. The way I structured the code means that all pointers are held in memory before any printing is done (in my system this meant 2MB of memory concumption, I don't know how high it can get), so it's something to be aware of.

However, if you think this is sound and something we should stick to (and also that the energy spent is worth it), I would be interested in rewriting most printing modes to go through the functions I wrote. This would allow us to use arbitrary ordering in all of those that list packages, and would go together with implementing the formatting stuff. We could even have "default" formatting modes to keep the tool compatible with scripts people wrote.

bin/xbps-query/list.c
```diff
@@ -70,77 +70,184 @@ list_pkgs_in_dict(struct xbps_handle *xhp UNUSED,
...
+}
+
+bool filter_hold_pkgs(xbps_object_t obj)
+{
+	return xbps_dictionary_get(obj, "hold");
+}
+
+bool filter_repolock_pkgs(xbps_object_t obj)
+{
+	return xbps_dictionary_get(obj, "hold");
```
Duncaen, line 105: `repolock`

ericonr: good eye

ericonr:
Performance impact, not the best testing ever. Memory usage measurement done through valgrind.

```
xbps-query -m --orderby installed_size
```

- on glibc, with 114 manually installed packages, on an SSD, it takes around 16ms to run and uses up 2MiB of memory.
- on musl, with 2400 manually installed packages, on an HD, it takes around 500ms to run with caches dropped, 150ms on the second run, and up to 11MiB of memory.

Sounds ok to me, shouldn't be a bottleneck. Also, it's pretty much the same amount of memory and time that default xbps-query uses, without sorting. So I think it's ok :)

ericonr:
@Duncaen ping

Chocimier:
I would prefer to have implemented (to implement?) `--format`, and make use of `sort -k`

camoz:
I think for the flag name, `--order-by` reads better than `--orderby`.

PR #280, opened 2020-05-01
xbps.d.5: Expand ignorepkg description, fix typo.

flexibeast:


travankor:
Any context for this change?

Also, this behaviour is the same for: `noextract`, `include`, `preserve`, `repository`, `virtualpkg`

flexibeast:
@travankor: Someone on IRC asked how to ignore multiple packages; the existing documentation isn't clear on how to do this. Apart from the existing text for `ignorepkg`, `xbps.d(5)` currently says:

     The configuration files can set multiple keywords that are pairs of keys and values, such as `key=value`.

which is ambiguous, and can be read as suggesting that multiple values can be supplied to a single `ignorepkg` instance, which is not the case.

Perhaps instead of modifying the `ignorepkg` description specifically, there should be text in the "Description" part stating the "multiple instances allowed, but one value per instance" constraint for  `ignorepkg` and the keywords you mentioned?

travankor:
That sounds like a good approach (rewording last paragraph in "Description").

Apparently, `/usr/share/xbps.d/xbps.conf` serves as an annotated example, but it's not 100% in sync with that latest xbps features.

ericonr:
We could mention the `/usr/share/xbps.d/xbps.conf` file more explicitly in that man page, perhaps.

flexibeast:
PR updated to reflect comments so far.

flexibeast:
Ping?

PR #281, opened 2020-05-01
xbps-install.1: Add further details about `-n`.

flexibeast:


travankor:
Another solution is to error out with -Sn and print a message.

flexibeast:
@travankor: Yes, i think an error message would be preferable, but i'd be interested to know if there are reasons why this might not be a good idea.

travankor:
Having an error message would be better IMO.

I tested `-Sn` and xbps doesn't do anything. However, `-Snu` seems like it performs `-nu`, which is wrong.

travankor:
Here's my alternative patch:
```
diff --git a/bin/xbps-install/main.c b/bin/xbps-install/main.c
index c626d40..1a36928 100644
--- a/bin/xbps-install/main.c
+++ b/bin/xbps-install/main.c
@@ -207,6 +207,11 @@ main(int argc, char **argv)
        if ((!update && !syncf) && (argc == optind))
                usage(true);
 
+       if (syncf && drun) {
+               printf("Can't use dry-run mode with 'xbps-install -S', please run `xbps-install -M`\n");
+               exit(EXIT_FAILURE);
+       }
+
        /*
         * Initialize libxbps.
         */
```

Duncaen:
I don't think adding conflicts for dry run makes sense, you use it to "test" commands without changing the system.

flexibeast:
@Duncaen: So you'd prefer to not have any error message? If so, would updating the man page as per this PR be okay? If not, i'll just close this.

travankor:
@Duncaen I think the issue is that dry-run with `xbps-install -S` might give the user the wrong idea since dry-run mode can't update the repository index. `xbps-install -Sn` is effectively `xbps-install -n`, while memory sync gives the closest result to what the user might have expected.

Admittedly, I'm no good with C, so I don't know if there's is a more technically correct way, or if all that needs to be done is a manual page update.

flexibeast:
Ping?

PR #283, opened 2020-05-02
Reword cache cleaning with dry-run enabled.

travankor:
Instead of incorrectly saying "Removed ${package}", rephrase to say "Would remove ${package}" when dry-run is enabled.

Based on comment https://github.com/void-linux/xbps/pull/281#issuecomment-622924113


Duncaen:
Dry run kind of loses its value if the output changes.

travankor:
Could dry run prepend or append "Dry Run" or an equivalent to the output? Because it was kind of confusing without an explicit notice.

PR #286, opened 2020-05-12
bin/xbps-repodb: add new tool to manage multiple repositories at once

Duncaen:
This currently only supports -p/--purge to remove obsolete binary
package files from repositories with the awareness that different
noarch package version can co-exist within a repository in
different architectures repository indexes.

### DIFF
--- a/bin/xbps-repodb/main.c
+++ b/bin/xbps-repodb/main.c
@@ -1,0 +1,122 @@
+/*-
+ * Copyright (c) 2020 Duncan Overbruck <mail@duncano.de>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT
+ * NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+ * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+#include <errno.h>
+#include <unistd.h>
+#include <getopt.h>
+
+#include <xbps.h>
+
+#include "defs.h"
+
+
+static void __attribute__((noreturn))
+usage(void)
+{
+	fprintf(stdout,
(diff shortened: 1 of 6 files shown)


bin/xbps-repodb/purge.c
```diff
@@ -0,0 +1,284 @@
...
+			struct arch *arch = calloc(1, sizeof (struct arch));
+			if (arch == NULL) {
+				perror("calloc");
+				exit(1);
+			}
+			if ((size_t)(d-name) >= sizeof arch->arch) {
+				xbps_error_printf("invalid repodata: %s\n", name);
+				exit(1);
+			}
+			strncpy(arch->arch, name, d-name);
```
pullmoll, line 93: `strncpy(3)` should use the size of the destination for `n`, i.e. `strncpy(arch->arch, name, sizeof(name));`, not a size depending on the source. I see you made sure it fits into `arch->arch` before but gcc or clang will probably complain?

pullmoll: Hm... no, they don't complain. Probably gcc is smart enough to *see* that this is ok.

Duncaen: This is used to copy only the architecture part from from `XXX-repodata`, n is used here as how many bytes to copy and not as limit, for the limit use case I would have used `xbps_strlcpy`.

pullmoll:
I have another use case. I accidentially just added `*.mips-musl.xbps` packages to my `x86_64-repodata`, because I forgot to prepend `XBPS_ARCH=mips-musl` to the command. DUH!
Could we make this tool remove (optionally?) wrong architecture packages from a repodata file?

Duncaen:
I think its better to not allow `xbps-rindex -a` to add different architectures packages anymore, I can't think of any usecase where this would be desirable and this would make this mode not really required.

Duncaen:
Hmm this shouldn't have been possible, `xbps-rindex -a` already checks the architecture: https://github.com/void-linux/xbps/blob/15941299ff677c8a17dbb7b73b3f811eba7cd69c/bin/xbps-rindex/index-add.c#L271-L276

pullmoll:
My bad. I didn't see the error message and for `*.noarch.xbps` it succeeded, which made me think it succeeded for all packages. Inspecting index.plist there is no `<string>mips-musl</string>` so everything is ok.

Duncaen:
My plan for `xbps-repodb` is to eventually move all the staging stuff to it to allow it to maintain the stage over multiple repositories and make `xbps-rindex` a tool for "simple" local repositories.

* `xbps-rindex` loses some of its functionality.
  * signing of packages moves to `xbps-repodb` (this is a task for published repositories).
  * `--clean` could stay to allow simple cleaning of stagedata.
  * `--remove-orphans` may stay for simple usecases, but `xbps-repodb --purge` is prefered. 
* `xbps-repodb` becomes the publishing tool for repositories.
  * it checks the stagedata and selectively moves sets of packages from each stagedata into the corresponding repodata when all dependencies and shared libraries can be satisfied.
  * sign repodata.
  * sign packages.
  * maybe maintain a new filesdb in the future?
  * maybe also maintain a man page dump?

`xbps-src` would not be changed it would continue to just use xbps-rindex to add new packages to *-stagedata.
`xbps-repodb` would be "privileged" ans has access to secret keys so it can sign repodata and packages.
It should be "fast" and non-disruptive to run it in a cronjob to have the secret keys owned by a user separate from `xbps-src`.

Chocimier:
Great that you found time to start working on this!

bin/xbps-repodb/main.c
```diff
@@ -0,0 +1,122 @@
...
+		{ "dry-run", no_argument, NULL, 'n' },
+		{ "verbose", no_argument, NULL, 'v' },
+		{ "version", no_argument, NULL, 'V' },
+		{ NULL, 0, NULL, 0 }
+	};
+	bool dry = false;
+	enum {
+		MODE_NIL,
+		MODE_PURGE,
+	} mode;
```
Chocimier, line 70: If it won't be initialized, reading it (in switch) is UB.

Chocimier:
Is it assured that for each stagedata there is repodata at same time?

bin/xbps-repodb/main.c
```diff
@@ -0,0 +1,122 @@
...
+}
+
+int
+main(int argc, char **argv)
+{
+	int c = 0, rv = 0;
+	struct xbps_handle xh;
+	const struct option longopts[] = {
+		{ "debug", no_argument, NULL, 'd' },
+		{ "dry-run", no_argument, NULL, 'n' },
```
st3r4g, line 61: missing "purge" long option here

bin/xbps-repodb/main.c
```diff
@@ -0,0 +1,122 @@
...
+	"MODE:\n"
+	" -p, --purge    Remove obsolete binary packages from repositories\n");
+	exit(EXIT_FAILURE);
+}
+
+int
+main(int argc, char **argv)
+{
+	int c = 0, rv = 0;
+	struct xbps_handle xh;
```
ericonr, line 58: Can't you do `struct xbps_handle xh = { };` instead of a memset below? It should initialize the whole struct to zero and is easier to follow.

ericonr: Well, this isn't ISO C. `struct xbps_handle xh = { 0 };` should work too.

PR #289, opened 2020-05-17
lib/transaction_files.c: fix leaving behind obsolete directories

### LINKED ISSUE #282: gcc update leaves directories of previous versions
As already mentioned in https://github.com/void-linux/xbps/issues/23#issuecomment-533779080, updating the gcc package leaves directories (including one file) of previous versions:

```
> tree /usr/lib/gcc/ -I 9.3
/usr/lib/gcc/
└── x86_64-unknown-linux-gnu
    ├── 8.2
    │   ├── include
    │   ├── install-tools
    │   └── plugin
    │       └── include
    │           └── config
    ├── 8.3
    │   └── include
    ├── 9.1
    │   └── include
    ├── 9.2
    │   └── include
    │       └── omp.h
    └── 9.3.0 -> 9.3

14 directories, 1 file
```

Duncaen:
Deleting obsolete directories is different from deleting obsolete files,
files need to be deleted by the first package that might change the file
into a different file type.
Directories need to be deleted by the last package that removes files
out of the directory to avoid ENOTEMPTY if another package has a file in
the given directory which is uninstalled later.

This fixes #282.

### DIFF
--- a/lib/transaction_files.c
+++ b/lib/transaction_files.c
@@ -443,7 +443,16 @@ collect_file(struct xbps_handle *xhp, const char *file, size_t size,
 		} else if (type == TYPE_DIR && item->old.type == TYPE_DIR) {
 			/*
 			 * Multiple packages removing the same directory.
+			 * Record the last package to remove this directory.
 			 */
+			if (idx < item->old.index || item->old.preserve)
+				return 0;
+			item->old.pkgname = pkgname;
+			item->old.pkgver = pkgver;
+			item->old.index = idx;
+			item->old.preserve = preserve;
+			item->old.update = update;
+			item->old.removepkg = removepkg;
 			return 0;
 		} else {
 			/*
@@ -777,9 +786,9 @@ xbps_transaction_files(struct xbps_handle *xhp, xbps_object_iterator_t iter)
 	while ((obj = xbps_object_iterator_next(iter)) != NULL) {
 		bool update = false;
 		/*
-		 * `idx` is used as package install index, to chose which
-		 * choose the first package which owns or used to own the
-		 * file deletes it.
+		 * `idx` is used as package install index, to choose which
+		 * choose the first or last package which owns or used to
+		 * own the file or directory deletes it.
 		 */
 		idx++;
 
--- a/lib/transaction_files.c
+++ b/lib/transaction_files.c
@@ -50,6 +50,8 @@ struct item {
 		const char *target;
 		uint64_t size;
 		enum type type;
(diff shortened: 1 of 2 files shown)


lib/transaction_files.c
```diff
@@ -777,9 +786,9 @@ xbps_transaction_files(struct xbps_handle *xhp, xbps_object_iterator_t iter)
 	while ((obj = xbps_object_iterator_next(iter)) != NULL) {
 		bool update = false;
 		/*
-		 * `idx` is used as package install index, to chose which
-		 * choose the first package which owns or used to own the
-		 * file deletes it.
+		 * `idx` is used as package install index, to choose which
+		 * choose the first or last package which owns or used to
+		 * own the file or directory deletes it.
```
ericonr, line 791: I'm not sure I understand this comment. The language is still a bit unclear.

ericonr: I understand idx is related to file removal, but the logic for the removal isn't very explicit.

Duncaen: I think the comment as it is is a bit out of scope for `idx`, I can add a toplevel comment to explain how everything works instead.

* `idx` is the packages index in the transaction, the order in which packages are removed and installed.
* for new files, symlinks and directories the first package has to remove the old file to avoid problems with file type changes, i.e. symlink to directory and to avoid that a package later in the transaction which previously owned the file removes the file which is now owned by the previously unpacked package.
* for obsolete files only the order of removal of directories is important, the directory has to be removed after the last file inside the directory has been removed, to avoid issues if multiple package updates/removals own files in that directory.

ericonr: I think copy-pasting your comment there would work well, thanks!

Perhaps mention how it works if the user also added a file in the directory themselves (although I don't think there are many cases where this is *supposed* to happen).

Duncaen: > Perhaps mention how it works if the user also added a file in the directory themselves (although I don't think there are many cases where this is supposed to happen).

* If the directory is just obsolete it will just do nothing about it and leave user file and directory behind.
* If the directory is supposed to be replaced with a regular file or a symlink, it needs to be able to delete directory, in this case it will count how many files are being removed and how many files are there. If there are more files than it can remove, xbps will abort the transaction before unpacking any packages. The whole file check happens after downloading and verifying binary packages (it needs the files list from the packages metadata) but before any other change is done to the system or the package database.

ericonr: Got it, I think. Thanks a lot!

tamikawilkinsontmp:
Will it work on package remove too?

PR #306, opened 2020-07-11
signature fixups

Duncaen:


lib/transaction_fetch.c
```diff
@@ -209,17 +209,41 @@ xbps_transaction_fetch(struct xbps_handle *xhp, xbps_object_iterator_t iter)
...
-			continue;
+			if (xbps_repo_pubkey(repo) == NULL) {
+				rv = errno;
+				xbps_error_printf("%s: failed to get public key: %s\n",
+				    repoloc, strerror(errno));
+				goto out;
+			}
+
+			/*
+			 * Download binary package and signature if either one
```
Chocimier, line 234: Enqueue downloading?

lib/transaction_fetch.c
```diff
@@ -209,17 +209,41 @@ xbps_transaction_fetch(struct xbps_handle *xhp, xbps_object_iterator_t iter)
...
+			 * of them don't exist.
+			 */
+			if (!xbps_remote_binpkg_exists(xhp, obj)) {
+				if (!fetch && !(fetch = xbps_array_create())) {
```
Chocimier, line 238: Average transaction do fetch package, just creating array on top will be not only more readable, but also faster.

lib/verifysig.c
```diff
@@ -82,26 +82,35 @@ bool
...
+	if (st.st_size != sizeof buf) {
+		xbps_error_printf("invalid signature file %s: size mismatch\n",
+		    sigfile);
+		(void)close(fd);
+		return false;
+	}
 
-	return val;
+	rd = read(fd, buf, sizeof buf);
+	if (rd == -1) {
```
Chocimier, line 105: `rd != sizeof buf`, and `errno ? errno : EIO` ?

ericonr: I guess `sigfile` is required to have a size of `512` bytes, right? 

It would be good if this could be a `#define` somewhere, and documented in comments. And I would just use `rd != SIGFILE_SIZE`, then.

PR #309, opened 2020-07-22
bin/xbps-create: fix incorrect assert variable

daltomi:


Chocimier:
How do you found it?

daltomi:
With `cppcheck`, but I don't remember if it indicated this directly.

Issue #301, opened 2020-07-03
Add EXAMPLE sections to man pages


ericonr:
This is something that might have been solved by the Rosetta stone docs page too, but I think we can fix it by having EXAMPLEs in all the man pages, for common use cases. Flag composition might not be obvious to users coming from other distros.

flexibeast:
i'm happy to work on this.

- [ ] xbps-alternatives
- [ ] xbps-checkvers
- [ ] xbps-create
- [ ] xbps-dgraph
- [ ] xbps-digest
  Unneeded?
- [ ] xbps-fbulk
- [ ] xbps-fetch
  Unneeded?
- [ ] xbps-install
- [ ] xbps-pkgdb
- [ ] xbps-query
- [ ] xbps-reconfigure
- [ ] xbps-remove
- [ ] xbps-rindex
- [ ] xbps-uchroot
- [ ] xbps-uhelper
  Covered in #305
- [ ] xbps-uunshare
- [ ] xbps.d

Chocimier:
I would love to see some nontrivial examples, like
- list installed self-build packages (I use `xbps-query --regex -p repository -s / | grep -v -e ' https://' -e ': /hostdir' | cut -d: -f1`
- create package depending on packages needed to build `ksysguard` on host
```
p=ksysguard
. $(xdistdir)/srcpkgs/${p}/template
xbps-create -n ${p}-build-deps-${version}_${revision} \
 -s "Build depends of ${p}" -A "$(xbps-uhelper arch)" \
 -D "$(echo $hostmakedepends $makedepends | sed 's/ />=0 /g')>=0" \
 $(mktemp -d)
```

flexibeast:
@Chocimier: `xbps-query` example added to #310, `xbps-create` example added in #314.

npranav7619:
Can I look into this ?

PR #310, opened 2020-07-23
xbps-query.1: Add EXAMPLES section.

flexibeast:
Partially addresses #301. 

Are the other examples that should be included?


ericonr:
Perhaps add an example for usage with the `--regex` option as well? Also, I guess we could include in this PR (somehow, I'm not sure about the phrasing) that the results shown with `-Rs` might include new versions of installed packages, so the output will show them as uninstalled.

flexibeast:
@ericonr: Thanks, PR updated.

Duncaen:
I feel like the examples that just use a single flag or `-RX` are redundant.

PR #311, opened 2020-07-23
xbps-pkgdb.1: Add EXAMPLES section.

flexibeast:
Partially addresses #301.

Are the other examples that should be included?


ericonr:
I would rather have other examples, I think. So like, one example of `xbps-pkgdb <package>` to check if package is fine; perhaps one of `xbps-pkgdb -a` to check all packages; and one `xbps-pkgdb -m <mode> <package>`, possibly with another command to revert the mode as well.

flexibeast:
@ericonr: So i could keep the 'hold' examples, remove the 'repolock' examples, and add in the other two examples you mentioned?

ericonr:
I think that works!

flexibeast:
@ericonr: Thanks, PR updated.

PR #314, opened 2020-07-27
xbps-create.1: Add EXAMPLES section.

flexibeast:


npranav7619:
Can i look into this

Duncaen:
I would prefer not to use `xbps-src` as example, the example from `xbps-create -h` would be fine.

PR #320, opened 2020-08-16
xbps-query: accept -- in command line.

### LINKED ISSUE #319: -- command line flag is broken for xbps-query
```
$ xbps-query -- gcc
[ information about gcc package ]
$ xbps-query -s -- gcc
[ help output of xbps-query ]
$ xbps-query --search -- gcc
[ help output of xbps-query ]
```

This is why running `xlocate <query>` without a local repo is outputting the xbps-query help. This was added to xlocate in https://github.com/leahneukirchen/xtools/commit/5edebea0a738117dde41077cd23756e79f4417e7, so sometime between then and now there was a regression.

ericonr:
Previously, `xbps-query -- gcc` worked, but `xbps-query -s -- gcc`
didn't, because options such as `--search` were configured as requiring
an argument. Instead, they should use the last parameter once all
options have been scanned.

Fixes #319 

I still need to test it thoroughly.

### DIFF
--- a/bin/xbps-query/main.c
+++ b/bin/xbps-query/main.c
@@ -77,7 +77,7 @@ usage(bool fail)
 int
 main(int argc, char **argv)
 {
-	const char *shortopts = "C:c:df:hHiLlMmOo:p:Rr:s:S:VvX:x:";
+	const char *shortopts = "C:c:dfhHiLlMmOop:Rr:sSVvXx";
 	const struct option longopts[] = {
 		{ "config", required_argument, NULL, 'C' },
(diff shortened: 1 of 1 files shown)


ericonr:
Oh, this breaks `xbps-query -s -dbg`, which works in the current setup because it forces the second one to be used as parameter. This makes it necessary to use `xbps-query -s -- -dbg`, which I don't think is what we want.

ericonr:
I should probably close this PR, unless we want to do lots of hacks around the command line part.

PR #328, opened 2020-08-30
tests/xbps/xbps-digest: create tests.

ericonr:
The `big_file` one takes a while, but I think it's worth having it, if we're moving to BearSSL.


Duncaen:
Not a fan of the big file test, any way we could make it optional?

ericonr:
I'm not sure, I'm mostly cargo culting the atf stuff. It takes like 8s on my PC, is that too much?

Duncaen:
Yes, writing 2GB each testsuite run is really not that nice.

Duncaen:
This is useless wear on disks and +8 seconds makes tests even more annoying to run.

ericonr:
I did try it with BearSSL and it didn't blow up, so I guess I can just leave a note to make sure to test `xbps-digest` with big files once necessary for the final transition. Will remove the big file one, then.

ericonr:
Pushed without big file.

ericonr:
@Duncaen anything holding this one?

PR #349, opened 2020-11-22
fetch/http: add missing case when treating status code.

### LINKED ISSUE #348: xbps-fetch behavior can differ when called with -v
As seen in https://github.com/void-linux/xbps/blob/05ff04a0fb86dfcec770260326a623b5a7d1f914/lib/external/fetch/http.c#L1067 , using libfetch in verbose mode leads to it parsing HTTP headers despite the actual status code on the URL.

What happens, as seen in [this gist](https://gist.github.com/the-maldridge/69ba4260e716f474fe47369f6abc6ab1), is that a fetch request that fails without `-v` may succeed with `-v`, since the headers will be parsed for a redirection header despite the status code.

What remains to find out is if this is an issue from SourceForge's servers or libfetch misinterpreting the status codes.

ericonr:
The HTTP_REDIRECT() macro already included this case, but the
switch-case with the return value of http_get_reply(conn) didn't. This
made it so that a server returning HTTP_TEMP_REDIRECT (307) would
generate a failure, unless libfetch was used in verbose mode, in which
case the headers would be parsed (in order to report errors), then the
'Location' header would be found and a new connection would be made.

Fixes #348

### DIFF
--- a/lib/external/fetch/http.c
+++ b/lib/external/fetch/http.c
@@ -1021,6 +1021,7 @@ http_request(struct url *URL, const char *op, struct url_stat *us,
 		case HTTP_MOVED_TEMP:
 		case HTTP_SEE_OTHER:
 		case HTTP_USE_PROXY:
+		case HTTP_TEMP_REDIRECT:
 			/*
 			 * Not so fine, but we still have to read the
 			 * headers to get the new location.


ericonr:
We should probably look at resyncing some changes from freebsd's version... This bug is fixed there (or never existed), and they also use `HTTP_PERM_REDIRECT`, which is currently completely unused here.

PR #352, opened 2020-12-05
Support running only specific checks in xbps-pkgdb

ericonr:
WIP because I have no idea how to construct the array to pass to `getsubopt`...

Using the example in https://man.voidlinux.org/getsubopt.3 complains about discarded qualifiers in `char *const available[]`, using the example in https://man.voidlinux.org/getsubopt.3p complains about the discarded qualifiers when passing `const char *available[]` to the function.

Since XBPS aims to build with `-Werror`, I need to find some way of satisfying the compiler.


hippi777:
hi there! :)

i took a deeper look at it, and other than the remaining issue, it looks all good to me! :)

Duncaen:
I don't think `-c` is a good flag, this is already used in most other tools for `--cachedir` and I could see that we might want to add this flag to `xbps-pkgdb` in the future.

I think it would generally fit better into the codebase to use an integer/bitmasks:

```
enum checks {
CHECK_FILES = 1 << 1,
CHECK_DEPENDENCIES = 1 << 2,
CHECK_ALTERNATIVES = 1 << 3,
CHECK_PKGDB = 1 << 4,
};
```

I think it makes sense to merge symlinks and files into one category.
"unneeded" is something internal and I think it would make more sense to call it pkgdb checks,
so this could be extended in the future to have more pkgdb checks without having to add more
flags. "rundeps" is an internal name, there are no different types of user facing dependencies
in xbps itself so I think calling it "depends" or "dependencies" makes more sense.

ericonr:
> I don't think -c is a good flag, this is already used in most other tools for --cachedir and I could see that we might want to add this flag to xbps-pkgdb in the future.

Hmm, good point. Suggestions for flag name? I think it would be nice to have a shortopt option for it.

> I think it would generally fit better into the codebase to use an integer/bitmasks:

Fair enough. It also simplifies initialization.

> I think it makes sense to merge symlinks and files into one category.
> "unneeded" is something internal and I think it would make more sense to call it pkgdb checks,
so this could be extended in the future to have more pkgdb checks without having to add more
flags. "rundeps" is an internal name, there are no different types of user facing dependencies
in xbps itself so I think calling it "depends" or "dependencies" makes more sense.

Indeed, the split may have been too granular. I will update it with your suggestions.

Thanks!

ericonr:
For now, I went with no shortopt.

I implemented your suggested changes.

PR #356, opened 2020-12-06
build: add --disable-werror option.

ericonr:
Makes it possible to override the -Werror flag.


Chocimier:
Isn't exporting `CFLAGS=-Wno-error` enough?

ericonr:
Yes, I somehow didn't think of it.

PR #359, opened 2020-12-12
lib: replace xbps_file_hash_check_dictionary with transaction_file lookup

Duncaen:
xbps_file_hash_check_dictionary was called for every file that is
getting unpacked, because the files list is an array it iterated over
the whole files array to find the matching file.
With a lot of files this is really slow and a lot of time was spend
in locking the proplib array and iterating over it.

Time from my Ryzen 3700X system with nvme disk updating
texlive-fontsextra with files 86375:

time xbps-install -y texlive-fontsextra
    6m34.61s real     6m27.71s user     0m03.95s system

And with this patch:

time xbps-install -y texlive-fontsextra
    0m08.40s real     0m07.34s user     0m00.98s system


lib/package_unpack.c
```diff
@@ -387,8 +390,9 @@ unpack_archive(struct xbps_handle *xhp,
 					}
 					rv = 0;
 				} else {
-					rv = xbps_file_hash_check_dictionary(
-					    xhp, binpkg_filesd, "files", buf);
+					xbps_dbg_printf(xhp, "%s: %s\n", entry_pname, file->sha256);
+					assert(file->sha256);
```
ericonr, line 394: Might be nice to make it clear that a `(null)` there is an error? Not sure.

`file->sha256 ? file->sha256 : "error..."`

include/xbps_api_impl.h
```diff
@@ -135,14 +155,16 @@ bool HIDDEN xbps_transaction_store(struct xbps_handle *, xbps_array_t, xbps_dict
 int HIDDEN xbps_transaction_init(struct xbps_handle *);
 int HIDDEN xbps_transaction_files(struct xbps_handle *,
 		xbps_object_iterator_t);
+void xbps_transaction_files_free(void);
 int HIDDEN xbps_transaction_fetch(struct xbps_handle *,
 		xbps_object_iterator_t);
 int HIDDEN xbps_transaction_pkg_deps(struct xbps_handle *, xbps_array_t, xbps_dictionary_t);
 
+struct transaction_file HIDDEN *
+xbps_transaction_file_new(struct xbps_handle *xhp UNUSED, const char *path);
```
ericonr, line 164:
```suggestion
xbps_transaction_file_new(struct xbps_handle *, const char *);
```

ericonr:
For the record, a good test case here seems to be updating papirus-icon-theme. It took 2m37s going from `20201001_1` to `20210201_1`, all of it inside `xbps-install`, not any hooks.

I will build this locally and see how better it is.

Duncaen:
This should greatly reduce the time needed, not sure why this example with more files is faster, maybe its just  a lot more symlinks or smaller files that would need to be checksumed.

For the implementation, I think I would rename a few things, `xbps_transaction_file_new` does not really represent what the function actually does and maybe it would make sense to call the structure xbps_file instead og xbps_transaction_file as the struct could be reused for other things in the future.
Might also make sense to move some of the fields like "bool update" back into the transaction_files item structure as some of those are not important outside of the context of finding conflicts in the transaction.

include/xbps_api_impl.h
```diff
@@ -135,14 +155,15 @@ bool HIDDEN xbps_transaction_store(struct xbps_handle *, xbps_array_t, xbps_dict
 int HIDDEN xbps_transaction_init(struct xbps_handle *);
 int HIDDEN xbps_transaction_files(struct xbps_handle *,
 		xbps_object_iterator_t);
+void xbps_transaction_files_free(void);
 int HIDDEN xbps_transaction_fetch(struct xbps_handle *,
 		xbps_object_iterator_t);
 int HIDDEN xbps_transaction_pkg_deps(struct xbps_handle *, xbps_array_t, xbps_dictionary_t);
 
+struct transaction_file *xbps_transaction_file_new(struct xbps_handle *, const char *);
+
```
ericonr, line 164: The naming is a bit confusing, in that `xbps_transaction_file_new` was added, but no `xbps_transaction_file_free`, and `xbps_transaction_files_free` was added, without a corresponding `_new`, and doesn't seem to be called at all.

Duncaen: Yes the _new does not actually create anything, it gives you a reference to a file struct that already exists from collecting all files in the transaction, "_new" because it returns the new file as opposed to the old file.

Not sure how this should be done exactly, something like `xbps_transaction_get_file` with either _new or _old or a parameter to choose which one you want.

Duncaen: `xbps_transaction_files_free` deletes the whole hashtable/tree, this should be called from somewhere after the transaction is done, maybe from `xbps_transaction_commit`.

lib/transaction_files.c
```diff
@@ -526,6 +508,8 @@ collect_file(struct xbps_handle *xhp, const char *file, size_t size,
 		item->new.update = update;
 		item->new.removepkg = removepkg;
 		item->new.target = target;
+		if (sha256)
+			item->new.sha256 = strdup(sha256);
```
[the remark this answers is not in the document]
Duncaen: Generally everything is zero initialized, IMHO not doing that by default when initializing the memory should only be done if you know you overwrite it immediately.

include/xbps_api_impl.h
```diff
@@ -95,6 +95,26 @@
...
+	TYPE_CONFFILE,
+} transaction_file_type_t;
+
+struct transaction_file {
+	const char *pkgname;
+	const char *pkgver;
+	char *sha256;
+	const char *target;
+	uint64_t size;
+	transaction_file_type_t type;
```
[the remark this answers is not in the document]
Duncaen: Maybe we need two struct a `xbps_transaction_file` and a `xbps_file` which is part of the former.
The idea would be to start using the `xbps_file` struct in places where we currently use the `xbps_dictionary` references to `files.plist` so that we gradually switch from dynamic plist types to actual structs which should help to avoid some issues at compile time through the type system.
I think `xbps_file` would be a simple way to start, doing this for packages would be the end goal, there are so many places that use `xbps_dictionary`s for packages and there are a lot of places that can be improved on the way to switching to c structs.

include/xbps_api_impl.h
ericonr: > Not sure how this should be done exactly, something like xbps_transaction_get_file with either _new or _old or a parameter to choose which one you want.

I think the parameter helps avoid duplicated logic, so I'm in favor of that.

PR #360, opened 2020-12-13
ci: Add GitHub actions CI.

ericonr:


the-maldridge:
@Duncaen I'm relatively satisfied with this, pcc seems flaky but I suspect that's just the nature of it.

PR #364, opened 2020-12-30
ci: enable pcc for all targets.

ericonr:
Thanks, leah :)


ericonr:
No idea why the test fails on glibc. Feel free to close the PR if you think adding pcc on glibc matters.

ericonr:
If anything, could be removing pcc from the test matrix.

PR #365, opened 2021-01-06
repo: return error code if no successful mirrors

### LINKED ISSUE #293: xbps-install -S exits with 0 despite ERROR (resolver failure)
Should be easily reproduced if you change or delete your /etc/resolv.conf :

```sh
bash-5.0# xbps-install -Su
[*] Updating `https://alpha.de.repo.voidlinux.org/current/musl/x86_64-musl-repodata' ...
ERROR: [reposync] failed to fetch file `https://alpha.de.repo.voidlinux.org/current/musl/x86_64-musl-repodata': Transient resolver failure
bash-5.0# echo $?
0
```

abenson:
Should preserve the "its okay for a mirror to fail" behavior, but still
catch in the case that it wasn't able to check any mirrors at all.

Closes #293.

### DIFF
--- a/lib/rpool.c
+++ b/lib/rpool.c
@@ -61,6 +61,7 @@ int
 xbps_rpool_sync(struct xbps_handle *xhp, const char *uri)
 {
 	const char *repouri = NULL;
+	int repocount = 0;
 
 	for (unsigned int i = 0; i < xbps_array_count(xhp->repositories); i++) {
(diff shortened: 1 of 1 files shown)


Duncaen:
As soon as you have one local repo, it wouldn't fail.

ericonr:
Is that necessarily bad? If you have configured local repos you should know what you're about; in the case you don't have any local repos, failing to fetch from any repository means something is definitely diverting from expected.

PR #368, opened 2021-01-25
Fix -Og build by generating correct compiler flags

Chocimier:
## Steps to reproduce:

With current master,

    export CFLAGS=-Og; make clean && ./configure && make --trace

## Expected behavior:

Build passes.

## Actual behavior:

Compilation fails because of incorrect compiler flags.

    make[2]: Entering directory '/home/piotr/github/xbps/bin/xbps-alternatives'
    gcc -DXBPS_SYSCONF_PATH=\"/usr/local/etc/xbps.d\" (cut) -Og -O2 -I. -I.. -I../include (cut) -I. -I.. -I../include (cut) -I. -I../.. -I../../include (cut) -I. -I../.. -I../../include (cut) -Og (cut)  -c main.c
    cc1: error: ../include: No such file or directory [-Werror=missing-include-dirs]
    cc1: error: ../lib/external/fetch: No such file or directory [-Werror=missing-include-dirs]
    cc1: error: ../lib/external/portableproplib: No such file or directory [-Werror=missing-include-dirs]
    cc1: error: ../lib/external/portableproplib/prop: No such file or directory [-Werror=missing-include-dirs]
    cc1: error: ../include: No such file or directory [-Werror=missing-include-dirs]
    cc1: error: ../lib/external/fetch: No such file or directory [-Werror=missing-include-dirs]
    cc1: error: ../lib/external/portableproplib: No such file or directory [-Werror=missing-include-dirs]
    cc1: error: ../lib/external/portableproplib/prop: No such file or directory [-Werror=missing-include-dirs]

Top makefile loads config.mk, then enters bin/ and loads config.mk again, what appends to previously constructed CFLAGS, then enters xbps-*, what loads config.mk again, but now with correct TOPDIR path, and appends to previous incorrect flags. Then makefile loads prog.mk, which loads config.mk again.

Fix makes config.mk to not use CFLAGS created in directories above.

In order to honor CFLAGS from environment, configure script now copies its env CFLAGS to config.mk. Flags from env in `make` step are thus ignored. This is what was likely intended, because configure script prints out own env CFLAGS. If this is correct, similar change for LDFLAGS and others should follow.

On ocassion, eliminate excessive includes, and replace `-include` directive with `include`, to make missing files early error rather than late error.

## Behavior after change

Build passes. One copy of flags from config.mk .

    gcc -DXBPS_SYSCONF_PATH=\"/usr/local/etc/xbps.d\" (cut) -O2 -I. -I../.. -I../../include (cut) -Og (cut) -c main.c

### DIFF
--- a/configure
+++ b/configure
@@ -192,11 +192,11 @@ fi
 [ -z "$DEBUG" ] && DEBUG=yes
 
 echo "CC =	$CC" >>$CONFIG_MK
-echo "CFLAGS +=	-O2" >>$CONFIG_MK
+echo "CFLAGS =	-O2" >>$CONFIG_MK
 
-echo "LDFLAGS +=  	-L\$(TOPDIR)/lib" >>$CONFIG_MK
+echo "LDFLAGS =  	-L\$(TOPDIR)/lib" >>$CONFIG_MK
 echo "CFLAGS += 	-I. -I\$(TOPDIR) -I\$(TOPDIR)/include" >>$CONFIG_MK
-echo "CPPFLAGS +=	-DXBPS_SYSCONF_PATH=\\\"${ETCDIR}\\\"" >>$CONFIG_MK
+echo "CPPFLAGS =	-DXBPS_SYSCONF_PATH=\\\"${ETCDIR}\\\"" >>$CONFIG_MK
 echo "CPPFLAGS +=	-DXBPS_SYSDEFCONF_PATH=\\\"${SHAREDIR}/xbps.d\\\"" >>$CONFIG_MK
 echo "CPPFLAGS +=	-DXBPS_VERSION=\\\"${VERSION}\\\"" >>$CONFIG_MK
 echo "CPPFLAGS +=	-DXBPS_META_PATH=\\\"${DBDIR}\\\"" >>$CONFIG_MK
--- a/tests/xbps/Makefile
+++ b/tests/xbps/Makefile
@@ -1,5 +1,6 @@
--include ../../config.mk
+TOPDIR = ../..
+include $(TOPDIR)/config.mk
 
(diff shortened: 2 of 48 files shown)


configure
```diff
@@ -192,6 +192,7 @@ fi
 [ -z "$DEBUG" ] && DEBUG=yes
 
 echo "CC =	$CC" >>$CONFIG_MK
+echo "CFLAGS =	" >>$CONFIG_MK
 echo "CFLAGS +=	-O2" >>$CONFIG_MK
```
ericonr, line 196:
```suggestion
echo "CFLAGS =	-O2" >>$CONFIG_MK
```

But maybe -O2 shouldn't be hardcoded here?

tests/xbps/Makefile
```diff
@@ -1,4 +1,4 @@
--include ../../config.mk
+include ../../config.mk
```
ericonr, line 1: `$(TOPDIR)`?

tests/xbps/libxbps/Makefile
```diff
@@ -1,4 +1,4 @@
--include ../../../config.mk
+include ../../../config.mk
```
ericonr, line 1: same

Chocimier:
Wait, configure was like that until very recently: https://github.com/void-linux/xbps/commit/9effec0c4e2f50342254d42458275edbe7fd0ee3 . Was that change actually needed?

Chocimier:
Above commit reverted in parts that carry flags from upper dirs.

@ericonr: done.

Duncaen:
Why is the include removed from some Makefiles?

Chocimier:
It is noop there, gets included by mk/prog.mk later on. Can resign from removal if you wish.

Duncaen:
Is fine in this case.

Chocimier:
Resolved by 51b39002e5aae532884bf4c20059341e369b1958, changes in makefiles are nitpicky.

Issue #29, opened 2018-11-10
xbps checks space available in / 


Menelkir:
I have a lvm setup with /, /usr and /var in different LVs. In my last install, I've tried to install a lot of things at the same time and xbps checked the space available in / (actually my / is 2Gb with 50Mb used, my /usr is 20Gb), so I was unable to install with a batch. If I install individually, it will pass, so it seems xbps is checking strictly on space available in /, isn't that bad but it's an issue if you use lvm with a different configuration (like mine).

Duncaen:
https://github.com/voidlinux/xbps/issues/93

xtraeme:
This is a known bug and hopefully it will be fixed soon :)

PR #371, opened 2021-01-28
xbps-install add -T option to skip disk space check

CorporalCrawdad:
Add an option to trust that there is enough disk space to continue transaction. This is a workaround for #29.

### DIFF
--- a/bin/xbps-install/main.c
+++ b/bin/xbps-install/main.c
@@ -53,6 +53,7 @@ usage(bool fail)
 	    " -h, --help                  Show usage\n"
 	    " -i, --ignore-conf-repos     Ignore repositories defined in xbps.d\n"
 	    " -I, --ignore-file-conflicts Ignore detected file conflicts\n"
+	    " -T, --trust-disk-space      Don't fail from insufficient disk space for transaction\n"
 	    " -U, --unpack-only           Unpack packages in transaction, do not configure them\n"
 	    " -M, --memory-sync           Remote repository data is fetched and stored\n"
 	    "                             in memory, ignoring on-disk repodata archives\n"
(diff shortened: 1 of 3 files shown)


Duncaen:
Not a fan of adding random flags for this, especially for something that should be fixed in a better way.
I think a checking if a environment variable would be better.

lib/transaction_prepare.c
```diff
@@ -169,21 +169,22 @@ compute_transaction_stats(struct xbps_handle *xhp)
...
-		xbps_dbg_printf(xhp, "%s: statvfs failed: %s\n", __func__, strerror(errno));
-		return 0;
-	}
-	/* compute free space on disk */
-	rootdir_free_size = svfs.f_bfree * svfs.f_bsize;
-
-	if (!xbps_dictionary_set_uint64(xhp->transd,
-				"disk-free-size", rootdir_free_size))
-		return EINVAL;
+	if (xhp->flags & XBPS_FLAG_IGNORE_DISK_SPACE) {
```
ericonr, line 172: Isn't this using the wrong logic?

bin/xbps-install/main.c
```diff
@@ -107,6 +108,7 @@ main(int argc, char **argv)
 		{ "help", no_argument, NULL, 'h' },
 		{ "ignore-conf-repos", no_argument, NULL, 'i' },
 		{ "ignore-file-conflicts", no_argument, NULL, 'I' },
+		{ "trust-disk-space", no_argument, NULL, 'T' },
```
ericonr, line 111: Trust doesn't feel like the correct word; also, if it's going to be a command line arg I think it should exclusively be a long opt; but Duncaen has expressed preference for an env var.

PR #373, opened 2021-02-07
Libfetch update

Duncaen:
Syncing most of the code with the FreeBSD version
and fixing some things about the keep-alive handling.
This should fix some problems 

I've been testing this with a bash script:

```
#!/bin/bash
#

export LD_LIBRARY_PATH=$PWD/lib
cmd=$PWD/bin/xbps-fetch/xbps-fetch

XBPS_UHELPER_CMD=xbps-uhelper
XBPS_TARGET_WORDSIZE=64
XBPS_WORDSIZE=64
XBPS_TARGET_MACHINE=x86_64

vopt_if() {
        :
}
vopt_conflict() {
        :
}
vopt_bool() {
        :
}
vopt_enable() {
        :
}
vopt_with() {
        :
}

for dir in ~/void-packages/srcpkgs/*; do
        [[ -h "$dir" ]] && continue
        unset arr
        unset distfiles
        . "$dir/template"
        args=(${distfiles})
        [[ ${#args} -eq 0 ]] && continue
        mkdir -p "/tmp/xbps-fetch/${pkgname}-${version}"
        cd "/tmp/xbps-fetch/${pkgname}-${version}"
        valgrind "$cmd" -dv "${args[@]}" 2>&1 | tee "/tmp/xbps-fetch/$pkgname.log"
        rv=$?
        if [[ $rv -ne 0 ]]; then
                echo "ERROR: $rv"
                mv "/tmp/xbps-fetch/$pkgname.log" "/tmp/xbps-fetch/$pkgname.error.log"
        else
                mv "/tmp/xbps-fetch/$pkgname.log" "/tmp/xbps-fetch/$pkgname.good.log"
        fi
done
```


ericonr (commented):
I had wanted to do this, thanks for getting to it :)

The part where they try to propagate errors from `close(fd)` seems weird, but they must have a reason for it.

From what I understood last I looked at FreeBSD libfetch, they implemented their own SOCKS5 handling. Is it compatible with the interface in ours?

PR #378, opened 2021-02-17
New feature : packagekit backend for xbps

pandom79:
This patch allows to [packagekit](https://github.com/pandom79/PackageKit/tree/xbps-backend/backends/xbps) to work correctly.
Basically, i had to making non static the variables in pkgdb.c and transaction_files.c.
If you are interested to that, i'd say to check it , to test it and if all ok (it works well for me), i'll make a PR
to the packagekit team to merge it.

### DIFF
--- a/include/xbps.h.in
+++ b/include/xbps.h.in
@@ -679,6 +679,39 @@ struct xbps_handle {
 	 * 	- XBPS_FLAG_* (see above)
 	 */
 	int flags;
+	/**
+	 * @var pkgdb_fd
+	 *
+	 * Pkgdb file descriptor
+	 */
+	int pkgdb_fd;
+	/**
+	 * @var pkgdb_map_names_done
+	 *
+	 * The boolean value representing if have been set pkgname for all packages in pkgdb
+	 *
+	 */
+	bool pkgdb_map_names_done;
+	/**
+	 * @brief hashtab
+	 *
+	 * hash table to look up files by path
+	 */
+	struct item_transaction *hashtab;
+	/**
+	 * @brief items
+	 *
+	 * Array of items representing the files contained in a package
+	 *
+	*/
+	struct item_transaction **items;
+	/**
+	* @brief itemsidx and itemssz
+	*
+	* These variable are used to sort items
+	*/
+	size_t itemsidx;
+	size_t itemssz;
 };
 
 void xbps_dbg_printf(struct xbps_handle *, const char *, ...) __attribute__ ((format (printf, 2, 3)));
--- a/lib/initend.c
+++ b/lib/initend.c
@@ -50,6 +50,16 @@ xbps_init(struct xbps_handle *xhp)
 
 	xbps_dbg_printf(xhp, "%s\n", XBPS_RELVER);
 
+	/* Initialize pkgdb_fd and pkgdb_map_names_done */
+	xhp->pkgdb_fd = -1;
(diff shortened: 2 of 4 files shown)


ericonr (commented):
Removing static globals seems like a fine idea to me. Could you try to cleanup the indentation changes? They make reviewing the diff quite a bit noisier.

Regarding uthash, I don't think we should expose it. It's internal to XBPS and should be considered an implementation detail... Maybe have a pointer to a structure containing the parts that use uthash?

Chocimier:
On packagekit plugin:
linking to libxbps.so is not good idea because it requires to rewrite xbps binaries.

Important features are missing in this incomplete rewrite, like reading same conf files, updating xbps first (that will be problematic when soname of xbps and some other packagekit dependency is bumped at same time).

Calling xbps binaries and parsing output, what eg. `octoxbps` do would be better approach, but still not supported, just like octoxbps is not.

From non-technical side: add copyright notice when copying whole functions.

pandom79:
> Removing static globals seems like a fine idea to me. Could you try to cleanup the indentation changes? They make reviewing the diff quite a bit noisier.
> 
> Regarding uthash, I don't think we should expose it. It's internal to XBPS and should be considered an implementation detail... Maybe have a pointer to a structure containing the parts that use uthash?

Yeah, it seems look much better now.

pandom79:
> Important features are missing in this incomplete rewrite, like reading same conf files, updating xbps first (that will be problematic when soname of xbps and some other packagekit dependency is bumped at same time).
> 

I decided to have a different configuration directory.
I forget to add xbps.conf and xbps-arch.conf because i had not problem during developing and neither at runtime.....;)
I could use the default xbps configuration directory and manage the repositories via install and remove repository package
(void-repo-debug, void-repo-multilib, etc etc).
My problem is not this, is useless change the code if the distro is not interested.
For my use case is enough.

> Calling xbps binaries and parsing output, what eg. `octoxbps` do would be better approach, but still not supported, just like octoxbps is not.
>

If the octoxbps developer invoked xbps binary and parsing the output,
we understand two important reasons :

1) That confirms that there are evident problems when someone wants to use xbps shared library. 
The static variables can also be used but they have not to have impact on important functionality.
The static variables in a shared library are not recommended, making , in this case, it unusable by a Gui.
   
2) Probably, the octoxbps developer had not time or cannot to find a better solution.
This patch will allow to him rewrite his Gui calling the Xbps API in more efficient and fast mode as well as it should be.

I'm wondering: 
What is the sence to have a shared library and you cannot use it in all application context ?!

> From non-technical side: add copyright notice when copying whole functions.
>

Sorry, i disagree with you


Anyway, at the moment seems that this patch resolve a problem for me, for Void Linux and for octoxbps developer.

meator:
[Is PackageKit still alive?](https://web.archive.org/web/20240826184625/https://blogs.gnome.org/hughsie/2019/02/14/packagekit-is-dead-long-live-well-something-else/)

meator:
I'm not requesting you to close this PR, it is a genuine question. Is PackageKit still active and are its frontends still being actively used?

pandom79:
> I'm not requesting you to close this PR, it is a genuine question. Is PackageKit still active and are its frontends still being actively used?

I close it because I am not on Void anymore and long time is passed so this code likely doesn't work anymore.
Regard your question, I'd not know....

PR #383, opened 2021-03-06
include: don't make xbps.h be generated from a template.

ericonr:
`@@VERSION@@` never existed in the header, so it's fair to assume it
isn't actually necessary.

---

I think the headers should be a bit more organized than this even; maybe have the `sed` actually do something, because as it stands we have a public header that depends on a bunch of compile time defines to even work properly.


ericonr:
Closed in favor of #416

PR #388, opened 2021-03-11
ci: migrate to openssl

Chocimier:


Chocimier:
Online test, not good. Will fix later: https://github.com/void-linux/xbps/blob/master/tests/xbps/libxbps/shell/obsoletefiles_test.sh#L474

Chocimier:
Fixed. Didn't find other missing rootdir flag.

Duncaen:
Those issues with the tests are really annoying even more that running tests with `./configure --prefix=/usr` completely breaks them, maybe there should be a `XBPS_ROOT` environment variable to avoid all those mistakes in scripts.

Issue #387, opened 2021-03-11
mtime mismatch errors with xbps-pkgdb


dkwo:
Are these expected, or do they signal an issue in my system?
Could you point to explanation how to fix?
```
doas xbps-pkgdb -u
doas xbps-pkgdb -a
ERROR: dbus: /etc/sv/dbus/check mtime mismatch (current: 1595220871, stored 1595218030)
ERROR: dbus: /etc/sv/dbus/run mtime mismatch (current: 1595220871, stored 1595218030)
ERROR: dbus: files check FAILED.
ERROR: bluez-alsa: /etc/alsa/conf.d/20-bluealsa.conf mtime mismatch (current: 1601894510, stored 1601869928)
ERROR: bluez-alsa: /etc/dbus-1/system.d/bluealsa.conf mtime mismatch (current: 1601894510, stored 1601869928)
ERROR: bluez-alsa: /etc/sv/bluez-alsa/run mtime mismatch (current: 1601894510, stored 1601869928)
ERROR: bluez-alsa: files check FAILED.
ERROR: pam: /etc/security/faillock.conf mtime mismatch (current: 1609727563, stored 1609726309)
ERROR: pam: /etc/security/namespace.init mtime mismatch (current: 1609727563, stored 1609726309)
ERROR: pam: files check FAILED.
ERROR: linux5.11: /boot/System.map-5.11.4_1 mtime mismatch (current: 1615177270, stored 1615177271)
ERROR: linux5.11: /boot/config-5.11.4_1 mtime mismatch (current: 1615177270, stored 1615177271)
ERROR: linux5.11: /boot/vmlinuz-5.11.4_1 mtime mismatch (current: 1615177270, stored 1615177271)
ERROR: linux5.11: files check FAILED.
```
Thanks.

Duncaen:
They are not important, you can fix them by force reinstalling the affected packages (`xbps-install -f`), but there is no reason to really do that unless there are hash mismatches.

dkwo:
I see, thank you.

PR #390, opened 2021-03-15
remove anything mtime related

Duncaen:
We don't rely on mtime anymore to detect files moved between packages, and mtime on fat filesystems is truncated by 2 seconds resulting in useless `xbps-pkgdb` warnings as seen in https://github.com/void-linux/xbps/issues/387.

### DIFF
--- a/bin/xbps-create/main.c
+++ b/bin/xbps-create/main.c
@@ -53,7 +53,6 @@
 
 struct xentry {
 	TAILQ_ENTRY(xentry) entries;
-	uint64_t mtime;
 	uint64_t size;
 	char *file, *type, *target;
 	char sha256[XBPS_SHA256_SIZE];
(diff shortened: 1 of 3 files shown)


ericonr:
I think mtime can be a useful shorthand to find out if anything changed permissions in a file. Maybe we could consider adding a check for it in `xbps-pkgdb`...

Duncaen:
Yes I think a permission check would be a lot more valuable than mtime checks.
The only issue currently with implementing this would be that this information is not part of `files.plist`, so you would have to have the package in the cache and would have to go through the whole tar archive to get the permissions out of it.

ericonr:
Why not add a new field to `files.plist` and have packages gradually fill those in? Could even add a fallback to the tarball, but I'm not sure it'd be worth it.

Duncaen:
I wouldn't be opposed to that, especially if we ever move make_dirs out of INSTALL/REMOVE scripts this would be a worthwhile addition to begin with.
Next step would probably optional user/group names.

PR #398, opened 2021-04-17
bin/xbps-checkvers: add -X flag to check and skip removed packages with existing revdeps

Duncaen:
This should make the list of removed packages safer to use by skipping removals of packages with existing revdeps.

In the x86_64 repo is currently one package which would be skipped, showing that it works.

```
$ ./bin/xbps-checkvers/xbps-checkvers -d -eX >/dev/null
skipping: libprotobuf22-32bit: protobuf-devel-32bit-3.11.4_1 depends on it
skipping: libprotobuf22-lite-32bit: protobuf-devel-32bit-3.11.4_1 depends on it
```

The  package (`protobuf-devel-32bit-3.11.4_1`) needs manual intervention, or some more logic in `xbps-checkvers` to notice that it should be removed (it was automatically generated, so there is no symlink, but since it was changed to a transitional package, the new `-32bit` package is not generated anymore).

### DIFF
--- a/bin/xbps-checkvers/main.c
+++ b/bin/xbps-checkvers/main.c
@@ -523,6 +523,9 @@ rcv_printf(rcv_t *rcv, FILE *fp, const char *pkgname, const char *repover,
 			p = strchr(rcv->fname, '/');
 			fwrite(rcv->fname, p ? (size_t)(p - rcv->fname) : strlen(rcv->fname), 1, fp);
 			break;
+		default:
+			fprintf(stderr, "ERROR: invalid format string\n");
+			exit(EXIT_FAILURE);
 		}
 	}
 	fputc('\n', fp);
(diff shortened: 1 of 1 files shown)


Duncaen:
@Chocimier

Chocimier:
From past irc disscussion:

> Chocimier> base-system depends on [virtual] awk, then we drop gawk and nawk, and make base-system [not depend on awk]
Chocimier> gawk and nawk will be removed before base-system is rebuild
duncaen> base-system depends on a specific awk implementation
duncaen> but I guess I can see the issue
duncaen> not planning to run it automatically after builds
duncaen> just to cleanup now and then instead of having to go through the list on 3 servers and delete things by hand
Chocimier> ship it then

> Chocimier> [repodb]
duncaen> at least for install/updates I've found SAT solvers are not the best solution
duncaen> error messages are important and there are certain solutions that are prefferable
duncaen> and I guess a SAT solver might be a lot better at solving the whole database as compared to just a subset of things where you need to come up with one solution and not "there is one or more solutions"

Duncaen:
Wouldn't repodb make this obsolete?

Chocimier:
Yes it would, instead of `xbps-checkvers -X | xargs rm; xbps-rindex -c` one could do `xbps-repodb; xbps-rindex -r`.

PR #402, opened 2021-05-12
Factor walk_dir into a library function

ericonr:
Also introduce `xbps_utils.h` header, because otherwise we'd have had to include `dirent.h` in `xbps_api_impl.h`, and that header is pretty full already.


lib/walk_dir.c
```diff
@@ -0,0 +1,83 @@
...
+			break;
+
+		cleanup:
+			free(list[i]);
+		}
+
+	}
+	for (; i >= 0; i--) {
+		free(list[i]);
+	}
```
ericonr, line 80: I'm not sure this doesn't double free, should check first before merging.

lib/walk_dir.c
```diff
@@ -0,0 +1,83 @@
...
+	int rv, i, count;
+	struct dirent **list;
+	char tmp_path[PATH_MAX] = { 0 };
+	struct stat sb;
+
+	count = scandir(path, &list, NULL, alphasort);
+	for (i = count - 1; i >= 0; i--) {
+		if (strcmp(list[i]->d_name, ".") == 0 || strcmp(list[i]->d_name, "..") == 0)
+			goto cleanup;
+		if (strlen(path) + strlen(list[i]->d_name) + 1 >= PATH_MAX - 1) {
```
Chocimier, line 50: `>`, as terminating null is already subtracted from PATH_MAX?

lib/walk_dir.c
Chocimier: If you refer to breaking out of previous loop: it is not.

Chocimier (approved):

PR #403, opened 2021-05-17
Minimize autoupdate transaction

Chocimier:


lib/transaction_ops.c
```diff
@@ -53,100 +53,120 @@
...
-			ttype = XBPS_TRANS_UPDATE;
+			*ttype = XBPS_TRANS_UPDATE;
 		}
 	}
 
+	return 0;
+}
+
+static int
+trans_find_pkg(struct xbps_handle *xhp, const char *pkg, bool force)
```
ericonr, line 152: maybe assert these too?

lib/transaction_ops.c
```diff
@@ -53,100 +53,120 @@
...
+	assert(ttype);
+	assert(pkg_pkgdb);
+	assert(pkg_repod);
 
 	/*
 	 * Find out if pkg is installed first.
 	 */
 	if (xbps_pkg_name(buf, sizeof(buf), pkg)) {
-		pkg_pkgdb = xbps_pkgdb_get_pkg(xhp, buf);
+		*pkg_pkgdb = xbps_pkgdb_get_pkg(xhp, buf);
```
ericonr, line 71: Maybe avoid so many pointer changes by using an intermediate variable and copying from it near the `return 0` line? Not sure it's worth it, but would minimize the diff at the expense of always remembering to copy things out at exit time.

lib/transaction_ops.c
```diff
@@ -53,100 +53,120 @@
...
+static int
+trans_find_pkg(struct xbps_handle *xhp, const char *pkg, bool force)
+{
+	xbps_dictionary_t pkg_pkgdb = NULL, pkg_repod = NULL;
+	xbps_object_t obj;
+	xbps_array_t pkgs;
+	pkg_state_t state = 0;
+	xbps_trans_type_t ttype;
+	const char *repopkgver, *pkgname;
+	int rv = 0;
```
ericonr, line 160: Doesn't need to be initialized.

lib/transaction_ops.c
```diff
@@ -53,100 +53,120 @@
...
+	return 0;
+}
+
+static int
+trans_find_pkg(struct xbps_handle *xhp, const char *pkg, bool force)
+{
+	xbps_dictionary_t pkg_pkgdb = NULL, pkg_repod = NULL;
+	xbps_object_t obj;
+	xbps_array_t pkgs;
+	pkg_state_t state = 0;
```
ericonr, line 157: `state` doesn't need to be initialized either, but idk if the compiler might complain with `maybe-uninitialized`; we'd have to make sure `xbps_pkg_state_installed` always initializes correctly, which seems like the right thing to do.

Chocimier:
Done, with initialization of `state` left.

PR #405, opened 2021-05-31
WIP : New feature : new xbps cache management

pandom79:
Xbps cache management

Since I had to make a terrific downgrade of a package, I decided to implement a way to keep the previous package in the cache.
Unfortunately, i have a bad habit to clean always the cache!!!
First of all, I changed the default behaviour, now it always will keep at least one previous version respect to current version.
If you want to remove all previous versions, you will have to do it via **'-k' or '--keep-cache'** option setting it to '**0**'.
In this case, will be show a warning before to proceed.
To be honest, this patch could be implement in various ways but I think I chose the way with less impact possible on the existing code.
Now I'm testing it and seems to work nicely.

Note: at the moment, the "keep_cache" parameter, after a first step, is set as static global variable but I'd prefer
pass it as parameter to the related functions.
Maybe, we could also think to transfer this functionality to the remove obsolete packages management in xbps-rindex.


bin/xbps-remove/clean-cache.c
```diff
@@ -32,10 +32,84 @@
...
+
+	if (xbps_pkg_name(pkgname, sizeof(pkgname), pkgver)) {
+		len = strlen(pkgname);
+		/* Building patterns */
+		pattern = xbps_string_create_cstring(pkgname);
+		assert(pattern);
+		xbps_string_append_cstring(pattern, "-");
+		xbps_string_append_cstring(pattern, "*.xbps");
+
+		assert(glob(xbps_string_cstring_nocopy(pattern), 0, NULL, &fake_results) == 0);
```
ericonr, line 66: You shouldn't have expressions with side effects inside assertions.

bin/xbps-remove/main.c
```diff
@@ -158,7 +173,7 @@ remove_pkg(struct xbps_handle *xhp, const char *pkgname, bool recursive)
 int
 main(int argc, char **argv)
 {
-	const char *shortopts = "C:c:dFfhnOoRr:vVy";
+	const char *shortopts = "C:c:dFfhH:nOk:oRr:vVy";
```
ericonr, line 176: Why did you add new stuff?

ericonr (commented):
I'm against changing the default behavior.

Checking if something is a number should be a general function, IMO.

So the idea here is for people to be able to clear their caches due to disk space but at the same time still keep *some* version of packages?

bin/xbps-remove/clean-cache.c
pandom79: That glob() function must forcely returned something. For this reason i assert that. Anyway, i can handle the result in different way.

bin/xbps-remove/main.c
pandom79: Basically, **keep_cache** parameter represents the number of previous versions of the packages that we want to keep in the cache making easier an eventual downgrade. Please ignore 'H' option, you only consider the new 'k' option. Copy and paste from my library in which i have more stuff.

pandom79: The idea is to have more control when we remove the obsolete packages from cache or local repository via "keep" option in both xbps-remove and xbps-rindex command. Currently, we always remove all previous versions of packages in both cache and local repository.

pandom79:
I just find out a bug in this solution. 
The order provided by glob function is not enough.
Before real results cycle i'll have to correctly order them by the version number. I'm working about that...

pandom79:
The main problem was the order returned by glob function.
The problem has occurred for the google-chrome package which versions are a bit complicated.
For the tests, i had three versions of google-chrome.

```
google-chrome-90.0.4430.77.1_1.x86_64.xbps
google-chrome-90.0.4430.212.1_1.x86_64.xbps
google-chrome-91.0.4472.77.1_1.x86_64.xbps
```
The order returned from glob function is the same if you run :
`ls google-chrome-*.xbps`
The output is :
```
google-chrome-90.0.4430.212.1_1.x86_64.xbps
google-chrome-90.0.4430.77.1_1.x86_64.xbps
google-chrome-91.0.4472.77.1_1.x86_64.xbps
```
This order follows the path name not the version name.

**google-chrome-90.0.4430.212.1_1.x86_64.xbps** package should follow **google-chrome-90.0.4430.77.1_1.x86_64.xbps**
In many case it worked but here, there is a bug.

According the 'ls' command guide https://www.gnu.org/software/coreutils/manual/html_node/Version-sort-overview.html
the -v option seems to be correct to use, in facts, it works nicely IMO.
The output is:

```
google-chrome-90.0.4430.77.1_1.x86_64.xbps
google-chrome-90.0.4430.212.1_1.x86_64.xbps
google-chrome-91.0.4472.77.1_1.x86_64.xbps
```

As you can see, it correctly follows the versions.

About the errors management, in this case, there are more errors to handle,
for this reason i use the assertions, if someone fails we'll have an error description in debug mode.

Chocimier:
Shelling out to ls is not an option, not only because it is not portable and breaks xbps-static's promise, but also it don't follow version order of xbps. Use xbps_cmpver to compare versions. Think also how ordering versions should interact with reverts.

Processing all files in directory for every package is too slow. xbps-rindex -r is called on builders, where there are tens of thousands of packages.

pandom79:
> Shelling out to ls is not an option, not only because it is not portable and breaks xbps-static's promise, but also it don't follow version order of xbps. Use xbps_cmpver to compare versions. Think also how ordering versions should interact with reverts.

Basically, the previous solution was good only that i'll have to order the result array with xbps_cmpver.
The correct order is essential for this patch to determine the correct package to remove according 'keep' parameter value.

> Processing all files in directory for every package is too slow. xbps-rindex -r is called on builders, where there are tens of thousands of packages.

For local use i didn't see much slowness (i didn't note the difference respect before) but if xbps-rindex -r is called on builders where there are many tens of thousands packages.....I start to think that this solution cannot be used by Void.
I couldn't imagine this otherwise I'd have not proposed that.
On the other side, i don't see another solution. For each package which has to be removed, we have to necessarily computing how many package are there to determine which packages must be remove.

pandom79:
It correctly sort the packages by version but, one more time, i don't see slowness problem locally.
I'd want to use qsort() to do that but seems that it doesn't work with xbps_array_t.
Moreover, about "revert", could you be more precise?
I only see that the packages with a revert key defined are ignored from is_removable_packages().

Could be an example the btrfs-progs package.
I only have three versions

```
btrfs-progs-5.9_1.x86_64.xbps
btrfs-progs-5.11.1_1.x86_64
btrfs-progs-5.12.1_1.x86_64.xbps
```

but, at most, it only deletes btrfs-progs-5.9_1.x86_64.xbps because
they are present in the index and thus always excluded seeing the currently management.

```
btrfs-progs
pkgver btrfs-progs-5.11.1_1
reverts 5.12.1_1
```

pandom79:
Sorry, i prefer qsort().
Please note that i can't make the cast in binpkgcmp() function
to get correctly the values.
I remove "cast-qual" flag and it works.
If you wish keep that flag, tell me how i can make the cast.

PR #407, opened 2021-06-06
lib/conf.c: strip whitespace after value in config

the-eater:


lib/conf.c
```diff
@@ -224,8 +224,10 @@ parse_option(char *buf, const char **keyp, char **valp)
 	while (isblank((unsigned char)*value))
 		value++;
 
-	/* eat final newline */
-	value[strlen(value)-1] = '\0';
+	if (isspace((unsigned char)value[strlen(value)-1])) {
+		/* eat final newline */
+		value[strlen(value)-1] = '\0';
+	}
```
ericonr, line 230: I wonder if we could instead eat all whitespace from the end of the line?

And I think in this case it would be cleaner to store the string length in a variable instead of calling `strlen` twice. Why did you add the `unsigned char` cast as well?

ericonr: Disregard my question about the cast, though the need for it is ugh :p

the-eater:
@ericonr processed your feedback whoo

lib/conf.c
```diff
@@ -224,8 +224,14 @@ parse_option(char *buf, const char **keyp, char **valp)
...
-	/* eat final newline */
-	value[strlen(value)-1] = '\0';
+	end = strlen(value);
+	/* eat trailing spaces, end - 1 here because \0 should be set -after- the first non-space
+	 * if end points at the actual current character, we can never make it an empty string
+	 * because than end needs to be set to -1, but end is a unsigned type thus would result in underflow */
+	while (end > 0 && isspace((unsigned char)value[end - 1]))
+		end--;
+
+	value[end] = '\0';
```
ericonr, line 234: This will be an out of bounds access if there is no newline:

```
value[end-1] will be '\0'
end won't be decremented
value[end] = '\0' will access beyond the string
```

ericonr: Not out of bounds, it's touching the null terminator.

PR #416, opened 2021-06-27
xbps.h: always populate XBPS_VERSION and XBPS_GIT.

ericonr:
Makes it so the generated header has the correct versions, even without
the CFLAGS passed by the build system.


configure
```diff
@@ -165,6 +165,13 @@ echo "PKGCONFIGDIR ?= $PKGCONFIGDIR" >>$CONFIG_MK
 echo "TESTSDIR ?= $TESTSDIR" >>$CONFIG_MK
 echo "DBDIR ?= $DBDIR" >>$CONFIG_MK
 
+if [ -d .git ] && command -v git >/dev/null; then
```
Chocimier, line 168: `$(git rev-parse --is-inside-work-tree 2>/dev/null) = true` instead of `-d .git` to handle worktrees?

ericonr: Ok!

PR #418, opened 2021-07-12
xbps-repodb: add --index

Chocimier:
Keeps repositories consistent within and between separate indices, includind delaying removal of package that is still depended on, and unstaging part by part.

Building by xbps-src with `xbps-rindex --stage -a` works, as xbps-install already use stagedata of local repos.

`xbps-rindex --remove` is independent, but actually useful now.

Repos are in pretty good shape already. Packages that don't index are mainly uncleaned removed packages, but there are also 32bit packages depending on arbitrarily `lib32disabled` packages, and few that become `broken` or `noarch` after their introduction.


classabbyamp:
I'm interested in writing the manpages for this once it gets merged :)

seems fine from a cursory read-through but I don't know enough C or enough about SAT solvers to give a real ok

PR #419, opened 2021-07-14
Link libxbps from binaries statically

Chocimier:
Allows to share code between binaries without exposing it to world.

Dynamic libxbps could be removed, as there are no known consumers.
Alternatively static only libxbps-private could be added and shared library could be trimmed to useful part (no xbps_archive_, xbps_mmap_file).

### DIFF
--- a/mk/prog.mk
+++ b/mk/prog.mk
@@ -49,11 +49,12 @@ endif
 $(BIN).static: $(OBJS) $(TOPDIR)/lib/libxbps.a
 	@printf " [CCLD]\t\t$@\n"
 	${SILENT}$(CC) -static $(OBJS) $(CPPFLAGS) -L$(TOPDIR)/lib \
-		$(CFLAGS) $(LDFLAGS) $(PROG_LDFLAGS) $(STATIC_LIBS) -o $@
+		$(CFLAGS) $(LDFLAGS) $(PROG_LDFLAGS) \
+		$(STATIC_LIBS) -o $@
 
-$(BIN): $(OBJS) $(TOPDIR)/lib/libxbps.so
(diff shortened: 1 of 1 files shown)


ericonr:
At least for now I'd rather have `libxbps-private.a`, otherwise we would bloat the binaries a tad. Have you checked the size difference?

The main improvement to size would be a multi call binary that's symlimked to the usual names. It would also greatly help xbps-static, which duplicates libssl and libcrypto in all binaries. But I think that's a discussion for later.

Re. consumers, I remember at least @the-eater mentioning doing things with libxbps via Rust, and @ArsenArsen doing via Python ctypes maybe? I'm sure about the former, not the latter.

ArsenArsen:
We've since decided using the CLI made more sense for our use case, since xbps is well scriptable. But I can confirm that I was, at one point, looking into libxbps.

the-eater:
I have been doing things with libxbps and preferably would like to the option to keep doing so? but mostly querying state of packages, which files it contains etc.

I talked about this before with @ericonr and one suggested option was putting the symbols in one of the binaries instead of libxbps which I am also fine with

DerNuntius:
I'm currently writing a libxbps based Objective-C library. It would be good to keep a shared version of libxbps.

0x5c:
I'm planning to make rust bindings for libxbps, focusing on access to package/repo info and transactions, with direct access to the data and without parsing xbps output.

Chocimier:
Some code is shared between binaries by implementing in `bin/` and linking just it into both, see https://github.com/void-linux/xbps/blob/c4ede2c7327be889926091de3886abf732c7fc93/bin/xbps-remove/Makefile#L6.

Therefore, linking whole library statically doesn't justify size increase.

PR #421, opened 2021-07-24
Print install and remove messages below transaction summary

Chocimier:
Makes messages actually visible, and allow to delay update in case manual steps required after update are inconvenient at given time.
As is, presence of messages breaks tools parsing xbps-install -n output, but I want that despite.
There are no install messages in xbps or octoxbps dependency tree, so system update won't be affected.

Pushed on top of static linking as publishing neither `xbps_cb_message` nor `show_package_msgs` used in xbps-install make sense.

Also changed formatting a bit, now it looks like:

```
qt5ct-1.2_1 install x86_64 https://alpha.de.repo.voidlinux.org/current 825203 226027
git-lfs-2.12.1_1 install x86_64 https://alpha.de.repo.voidlinux.org/current 15180953 7175847
=== rustup-1.24.1_1: install message ===================================
Please run rustup-init after initial install to use rustup.
========================================================================
=== qt5ct-1.2_1: install message =======================================
To enable qt5ct restart the X11 server for the changes to take effect.
========================================================================
=== git-lfs-2.12.1_1: install message ==================================
Run 'git-lfs install' to configure git to use git-lfs
========================================================================
```


bin/xbps-install/transaction.c
```diff
@@ -53,14 +53,17 @@ show_package_msgs(struct xbps_handle *xhp, xbps_object_iterator_t iter) {
...
 				break;
 			default:
-				xbps_cb_message(xhp, obj, "install-msg");
-				break;
+				{
+					const char *pkgname = xbps_string_cstring_nocopy(xbps_dictionary_get(obj, "pkgname"));
+					xbps_dictionary_t pkgdb_pkg = pkgname ? xbps_pkgdb_get_pkg(xhp, pkgname) : NULL;
+					xbps_object_t previous = pkgdb_pkg ? xbps_dictionary_get(pkgdb_pkg, "install-msg") : NULL;
+					xbps_cb_message(xhp, obj, "install-msg", previous);
+					break;
```
ericonr, line 65: Can you put a comment explaining this? I don't get the logic.

bin/xbps-remove/main.c
```diff
@@ -93,10 +95,12 @@ state_cb_rm(const struct xbps_state_cb_data *xscd, void *cbdata UNUSED)
...
 	case XBPS_STATE_SHOW_REMOVE_MSG:
-                printf("%s: pre-remove message:\n", xscd->arg);
-		printf("========================================================================\n");
-		printf("%s", xscd->desc);
-		printf("========================================================================\n");
+		chars_print = printf("=== %s: remove message ", xscd->arg);
+		if (chars_print < sizeof bar) {
+			printf("%s", bar + chars_print);
+		}
+		printf("\n%s", xscd->desc);
```
ericonr, line 102:
```
		if (chars_print < sizeof bar) {
			puts(bar + chars_print);
		}
                fputs(xscd->desc, stdout); // not sure if you like this one, otherwise keep the printf
```

ericonr (commented):
I think an important part missing here is an option to disable printing the info. I think having messages below the summary helps with discoverability, at least, but maybe we should consider printing a message to look above the messages for the summary, if any message was printed? Might help unaware users.

Unrelated: these install msgs should all be removed.

bin/xbps-install/transaction.c
Chocimier: Here install-msg of installed version of package is read from system pkgdb, and passed to xbps_cb_message.
Then xbps_cb_message don't print message if current and previous are same.
Is doc comment of xbps_cb_message best place to write it down?

Chocimier (commented):
> important part missing here is an option to disable printing the info

Why? I feel something is done wrong if this needs a toggle.

> maybe we should consider printing a message to look above

May be helpful when message take all screen, indeed.

Changed fprint to fputs.

Messages during transaction are still printed, and never skipped, but this can be changed to no more print during transaction, I think.

bin/xbps-install/transaction.c
ericonr: >  Then xbps_cb_message don't print message if current and previous are same.

Not sure how I feel about this... If someone misses an INSTALL.msg, they will never see it again.

>  Is doc comment of xbps_cb_message best place to write it down?

Yeah, I guess so. And comment this code block "get installed version of package, if any".

ericonr:
> Why? I feel something is done wrong if this needs a toggle.

You could have read them already and not want the spam, or to pipe the output through some filter/program (even if it's not advised), or something else I can't think of.

ericonr:
And I don't see why keep the messages *during* installation. The most I'd do is print them after as a block reminder.

Chocimier:
I'm not convinced about switch to disable. Config file don't fit, because you don't want to disable and forget. Flag don't fit, because existing tools won't use it, and adapted could use `| sed -n '/^=/q; p'` instead. What's left is env variable maybe.

> You could have read them already and not want the spam

This is solved by showing message just once, administrators of multiple systems are encouraged to skim over anyway.

> I don't see why keep the messages during installation.

removed

Mentioning packages list is added, but not really nice.

Chocimier:
Rebased, removed last line pointing to packages list and replaced static linking with exposing function, as agreed in other PR.

Duncaen:
I don't think breaking the `-n` output is good, what is the reason to print them there, would stderr be ok?

PR #427, opened 2021-09-16
cmpver: support epoch

sgn:
There're some packages that change version schema,
let's invent new field to address those packages.


Chocimier:
Looks useful.

- would be nice to have tests of xbps_binpkg_pkgver, xbps_pkg_version and similar with epoch
- epoch specified in middle (`1.2:3_4`) should be error, explicit zero epoch should be error as well, because some code assume string equality of equal versions

PR #431, opened 2021-10-04
xbps-rindex: add no-stash to disable stagedata

Gottox:
This PR adds a feature to disable repo staging when adding new packages

CC @the-maldridge @Duncaen

### DIFF
--- a/bin/xbps-rindex/xbps-rindex.1
+++ b/bin/xbps-rindex/xbps-rindex.1
@@ -33,6 +33,8 @@ This flag is only useful with the
 or
 .Em sign-pkg
 modes.
+.It Fl N -no-stash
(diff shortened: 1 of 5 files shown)


bin/xbps-rindex/xbps-rindex.1
```diff
@@ -33,6 +33,8 @@ This flag is only useful with the
 or
 .Em sign-pkg
 modes.
+.It Fl N -no-stash
+Do not initiate staging when inconstent shlibs are detected.
```
Chocimier, line 37: Please document what it does instead.

the-maldridge:
From discussion in IRC I think the preferred approach here is to use repodb to remove the need for staging entirely.

PR #436, opened 2021-12-04
xbps-query: list unavailable repositories in -L mode

Chocimier:
xbps-query: list unavailable repositories in -L mode

As documented in manpage. Example is when current user have no permissions to read local repo.


Duncaen:
What do you think about just looping over/opening `xhp->repositories` in `xbps-query` instead of using the rpool functions?

Chocimier:
Changed.

Duncaen:
Merged in https://github.com/void-linux/xbps/commit/ce4fd6a63cd88b0e7a2e0e6df6cfa58cadacb35c.

PR #438, opened 2021-12-08
xbps-install: exclude hold packages from total download size

Chocimier:


lib/transaction_prepare.c
```diff
@@ -100,10 +100,12 @@ compute_transaction_stats(struct xbps_handle *xhp)
 
 		if ((ttype != XBPS_TRANS_CONFIGURE) && (ttype != XBPS_TRANS_REMOVE) &&
```
Duncaen, line 101: wouldn't it be easier to add that here too?

Chocimier: Right.
I incorrectly assumed, that `xbps_dictionary_set_bool(obj, "download", true);` must not be ommited here for `--download-only`

PR #449, opened 2021-12-26
docs: Improve description of keepconf option

camoz:
keepconf is a boring option and is not needed in most situations. Make
this more clear in the documentation, e.g. that it only affects
*unchanged* configuration files, and that changed configuration files
are unaffected by this option.

### DIFF
--- a/data/xbps.conf
+++ b/data/xbps.conf
@@ -71,19 +71,20 @@
 #	preserve=/etc/file
 #	preserve=/etc/dir/*.conf
 
-## PRESERVING CONFIGURATION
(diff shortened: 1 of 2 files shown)


camoz:
I'm not sure if it's correct that I edited the nroff file directly. Seems like a PITA do edit nroff manually.

Duncaen:
Looks reasonable, I had my concerns already when this was added, the name of the option is already confusing.

https://github.com/void-linux/xbps/pull/214#issuecomment-583331682

camoz:
Yeah I don't like the name either. But too late now, I guess...

PR #452, opened 2021-12-27
Handle alternatives in transaction files

Duncaen:


Duncaen:
This should resolve all issues with switching files between real and alternatives (https://github.com/void-linux/xbps/pull/253).
Additionally files.plist is now loaded and added to the transaction files at the same time scripts are loaded from the binary package, instead of loading the binpkgs twice.

lib/transaction_files.c
```diff
@@ -652,124 +653,308 @@ collect_files(struct xbps_handle *xhp, xbps_dictionary_t d,
...
+		int r;
+
+		if (!providers) {
+			/*
+			 * There are no previous providers for the alternative
+			 * group, this package becomes the provider.
+			 */
+			providers = xbps_array_create();
+			xbps_array_add_cstring(providers, changer->pkgname);
+			if (!xbps_dictionary_set(alternatives, group, providers))
```
Chocimier, line 816: `xbps_dictionary_set_and_rel` prevents leak

lib/transaction_files.c
```diff
@@ -652,124 +653,308 @@ collect_files(struct xbps_handle *xhp, xbps_dictionary_t d,
...
-			filesd = xbps_archive_get_dictionary(ar, entry);
-			if (filesd == NULL) {
-				rv = EINVAL;
-				goto out;
+		/*
+		 * The alternative group already registers, add this package
+		 * as provider if it is not already part of it.
+		 */
+		for (unsigned int j = 0; j < xbps_array_count(providers); j++) {
+			const char *cur;
```
Chocimier, line 830: cur = NULL

lib/transaction_files.c
```diff
@@ -652,124 +653,308 @@ collect_files(struct xbps_handle *xhp, xbps_dictionary_t d,
...
+			xbps_array_get_cstring_nocopy(providers, i, &cur);
+			if (strcmp(cur, changer->pkgname) == 0) {
```
Chocimier, line 832: cur &&

lib/transaction_files.c
```diff
@@ -652,124 +653,308 @@ collect_files(struct xbps_handle *xhp, xbps_dictionary_t d,
...
-			rv = collect_files(xhp, filesd, pkgname, pkgver, idx,
-			    update, false, false, false);
-			xbps_object_release(filesd);
-			goto out;
 		}
-		archive_read_data_skip(ar);
+		if (!found)
+			xbps_array_add_cstring(providers, changer->pkgname);
+		if (!xbps_array_get_cstring_nocopy(providers, 0, &first)) {
+			/* XXX: does this need to be handled? */
```
Chocimier, line 840: This means
- xbps_array_add_cstring failed, or
- first elem of array is not a string, i.e. file is corrupted
thus should return error.
On the other hand, `*_add`, `*_set` calls are rarely checked for success in xbps.

lib/transaction_files.c
```diff
@@ -652,124 +653,308 @@ collect_files(struct xbps_handle *xhp, xbps_dictionary_t d,
...
+	 */
+	r = prune_alternative_groups(xhp, alternatives, &changer,
+	    new_pkg_alternatives, old_pkg_alternatives);
+	if (r < 0)
+		return r;
+	return 0;
+}
+
+static xbps_dictionary_t
+copy_alternatives(xbps_dictionary_t src)
```
Chocimier, line 942: is it `xbps_dictionary_copy_mutable`?

lib/transaction_files.c
```diff
@@ -849,6 +1036,7 @@ xbps_transaction_files(struct xbps_handle *xhp, xbps_object_iterator_t iter)
 			bool removepkg = (ttype == XBPS_TRANS_REMOVE);
 
 			xbps_dictionary_get_cstring_nocopy(pkgd, "pkgver", &oldpkgver);
```
Chocimier, line 1038: oldpkgver is unmodified on error, and assert below is passed unintialized value

Chocimier (commented):
Few nits. Had not read in depth yet.

xHyroM:
Any updatew about this?

PR #456, opened 2021-12-29
lib/rpool.c: remove ENOTSUP error from xbps_rpool_foreach

### LINKED ISSUE #454: xbps-install: Print error message when `dir` in `-r dir` does not exist
When running `xbps-install -S -r /path/does/not/exist <somepackage>` (and perhaps similar commands) an error message would be nice, saying that the path does not exist.

Duncaen:
This error is already ignored by most callers and where it wasn't it
masked the real error of packages simply not being found without
repositories.

Fixes #454.


Duncaen:
Looking at how `xbps_rpool_foreach` is used, the human readable errors reported from its failure as `Failed to initialize rpool` is somewhat wrong, the function can either fail for that reason or because the callback function returned an error.
I feel like this shouldn't happen, we should have a separate function to initialize the rpool and make `xbps_rpool_foreach` only return the callback functions return value, otherwise errors are impossible to identify at the caller.

PR #457, opened 2021-12-29
Fix crashes due to incorrectly registered alternatives

Chocimier:


Chocimier:
Has to be redone over #452.

PR #476, opened 2022-03-06
xbps-fetch: fix unreachable fetch result code and add tests

oopsbagel:
Adds test cases for xbps-fetch, including a test for the broken code.

Can someone please explain the long term goals surrounding `xbps-fetch` vs. `xbps-uhelper fetch`? They both perform the same function with slightly different implementations. I have a separate branch locally for adding similar tests to xbps-uhelper (which unmasks an error condition where it will return 0 if one of the downloads fails so long as the last one is succesful). I can push that instead, if xbps-fetch is meant to be deprecated, or later if it's intended to keep both tools.

I'm more than happy to change my testing methodology to confirm to project standards.


Duncaen:
Merged

Issue #447, opened 2021-12-26
Make output coherent: `...` vs ` ...` (leading space)


camoz:
Sorry for the nitpick, but I just noticed this. Sometimes there is a space before the `...`, sometimes not:

```
# xbps-install bmon

[...]

[*] Downloading packages
bmon-4.0_5.x86_64.xbps.sig: 512B [avg rate: 11MB/s]
bmon-4.0_5.x86_64.xbps: 42KB [avg rate: 260MB/s]
bmon-4.0_5: verifying RSA signature...

[*] Collecting package files
bmon-4.0_5: collecting files...

[*] Unpacking packages
bmon-4.0_5: unpacking ...

[*] Configuring unpacked packages
bmon-4.0_5: configuring ...
bmon-4.0_5: installed successfully.
```

PR #479, opened 2022-03-26
Don't use spaces before ellipses

mhmdanas:
Fixes https://github.com/void-linux/xbps/issues/447.

### DIFF
--- a/configure
+++ b/configure
@@ -147,7 +147,7 @@ if [ -z "$OS" ]; then
 	KERNEL=${REST%%-*}
 fi
 
-echo "Configuring xbps for ... $OS"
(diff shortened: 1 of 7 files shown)


configure
```diff
@@ -648,15 +648,15 @@ fi
 #
 if [ -n "$BUILD_API_DOCS" ]; then
 	echo "Building API documentation via doxygen and graphviz."
-	printf "Checking for doxygen ..."
+	printf "Checking for doxygen... "
```
mhmdanas, line 651: I took the liberty of adding what seemed to be a missing space here, hope you all don't mind.

PR #487, opened 2022-05-03
configure: make DEBUG able to be disabled

tianyuanhao:
There was no way to set `DEBUG` to `no` before.


tianyuanhao:
request review @Duncaen

PR #497, opened 2022-07-02
lib/fetch: reuse http connection

Chocimier:
With HTTP 1.1 persistent connection is default and Connection
header is not sent.
Before patch, for every file, including 512b signature, there was
done dns query, tls handshake etc.

### DIFF
--- a/lib/fetch/http.c
+++ b/lib/fetch/http.c
@@ -424,7 +424,7 @@ http_cmd(conn_t *conn, const char *fmt, ...)
  * Get and parse status line
  */
 static int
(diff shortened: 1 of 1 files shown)


Duncaen:
I guess this was never noticed because we had plain nginx before which always sends `Connection: keep-alive`.

https://alpha.de.repo.voidlinux.org/current/ sends it, https://repo-default.voidlinux.org/current/ doesn't.

Haagen-Dazs:
It would be awesome if we could also get a Static XBPS with this included.

PR #501, opened 2022-07-30
xbps-query: add message when no packages have been found

### LINKED ISSUE #494: Add an error message saying that no packages were found to xbps-query
The current behavior of xbps-query is to simply not display any output if packages aren't found, it would be useful to receive a message to standard error stating that no packages matched the query (at least in `-v[erbose]` mode) to clarify that the query actually completed.

OfekShochat:
closes #494

### DIFF
--- a/bin/xbps-query/search.c
+++ b/bin/xbps-query/search.c
@@ -58,6 +58,11 @@ print_results(struct xbps_handle *xhp, struct search_data *sd)
 	const char *pkgver = NULL, *desc = NULL;
 	unsigned int align = 0, len;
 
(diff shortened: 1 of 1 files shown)


crtxcr:
Just my opinion as another user: 

While I too would like to see a message, this PR one is incomplete because:

1) misses a newline at the end of the printf
2) does not apply to -f and is therefore, inconsistent
3) prints to stdout, not stderr

I would argue it should also be shown without -v.

JamiKettunen:
No idea, you'd have to get an actual review from a Void Linux maintainer, in particular @Duncaen seems to be the one merging on this repo.

Issue #519, opened 2022-11-18
Some strange bug where xbps-rindex cant access a path


MiloLug:
So I just tried to rindex my xbps-src's hostdir and I got left in complete confusion:

```bash
xbps-rindex -r hostdir/    # OK (though it's not even a repo dir, but I wanted to try)
xbps-rindex -r hostdir/binpkgs/    # not OK
xbps-rindex: failed to open hostdir/binpkgs/: No such file or directory
xbps-rindex -r "hostdir/binpkgs/"    # not OK
xbps-rindex: failed to open hostdir/binpkgs/: No such file or directory

cd hostdir/binpkgs/
xbps-rindex -r .    # OK
```

And the most interesting part here, when it works even with unreal path:
```bash
xbps-rindex -r jadhfladjfhaldjfhlajfhad    # OK
```

I just assumed it's some permissions problem, but it just refuses to work with the given path.

hostdir stats:
```
drwxrwxrwt  5 liava liava  4096 Sep 24 18:43 hostdir
```

binpkgs stats:
```
drwxrwxrwt  3 liava liava  4096 Nov 18 06:20 binpkgs
```

classabbyamp:
can you post a log for `xbps-rindex -d -r hostdir/binpkgs`?

MiloLug:
Ohh my, sorry.. I forget it requires absolute paths.
Here are the logs. But maybe I should close the issue
```
[DEBUG] XBPS: 0.59.1 API: 20200221 GIT: UNSET
[DEBUG] Processing configuration directory: /etc/xbps.d
[DEBUG] Parsing configuration file: /etc/xbps.d/10-ignore.conf
[DEBUG] Added ignored package: linux-firmware-amd
[DEBUG] Added ignored package: linux-firmware-broadcom
[DEBUG] Added ignored package: linux-firmware-network
[DEBUG] Added ignored package: ipw2100
[DEBUG] Added ignored package: ipw2200
[DEBUG] Processing system configuration directory: /usr/share/xbps.d
[DEBUG] Parsing configuration file: /usr/share/xbps.d/00-repository-main.conf
[DEBUG] [repo] `https://repo-default.voidlinux.org/current' stored successfully
[DEBUG] /usr/share/xbps.d/00-repository-main.conf: added repository https://repo-default.voidlinux.org/current
[DEBUG] Parsing configuration file: /usr/share/xbps.d/10-repository-nonfree.conf
[...]

PR #521, opened 2022-11-18
lib/repo.c: add user-facing error message when repodata doesn't exist

classabbyamp:
Instead of just cryptically exiting 0 when opening a non-existant repodata or stagedata, with the only message existing in the debug output, show an error to the user on stderr.

mentioned in #519


classabbyamp:
should this PR be expanded to add user-facing messages in other places where the exit condition is cryptic (i.e., just an exit code)?

Duncaen:
Should probably use `xbps_error_printf` for consistency.

PR #522, opened 2022-12-01
CI: migrate from LGTM to Github CodeQL

classabbyamp:
As of today, LGTM will [no longer be reviewing pull requests](https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/), and will shut down fully on 16 December. Github now runs the same service built-in as CodeQL, and integrates it into the [security tab](https://github.com/void-linux/xbps/security) of the repo.

I've replicated the LGTM setup to the best of my knowledge.

cc: @void-linux/xbps-developers 

(also includes a couple improvements to the build/test workflow I noticed)

### DIFF
--- a/.github/codeql/codeql-config.yaml
+++ b/.github/codeql/codeql-config.yaml
@@ -1,0 +1,5 @@
+queries:
+  - uses: security-extended
+  - uses: security-and-quality
+paths-ignore:
+  - tests/**
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,48 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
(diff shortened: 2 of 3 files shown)


.github/codeql/codeql-config.yaml
```diff
@@ -0,0 +1,5 @@
+queries:
+  - uses: security-extended
+  - uses: security-and-quality
```
classabbyamp, line 3: not sure what queries LGTM was using, but these built-in but non-default ones seemed useful

.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,48 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
```
classabbyamp, line 7: should this only run on PRs/pushes that touch code?

PR #533, opened 2023-02-09
bin/xbps-uhelper: add longopts, manpage

classabbyamp:
- bin/xbps-uhelper: add longopts for -CdrV
- bin/xbps-uhelper: add manpage
- data/_xbps: add xbps-uhelper getname, getversion to completions

I started (and mostly finished) this manpage before looking at the open PRs and seeing #305, and seeing as that PR has sat for a year and a half now, I'm submitting my alternative take on the uhelper manpage.

It's mostly the same, with some minor formatting differences, and a couple additional sections.

I did not include `xbps-uhelper fetch` and `xbps-uhelper digest` in the manpage because they are replaced by `xbps-fetch` and `xbps-digest` (https://github.com/void-linux/xbps/blob/master/NEWS#L357-L359).

closes #305

PR #536, opened 2023-02-09
bin/xbps-uhelper: allow multiple arguments for many actions

classabbyamp:
- allow multiple arguments for many actions

Only implemented where it makes sense to, i.e., not `cmpver`, `pkgmatch`, `arch`, `getsystemdir`, `digest`, and `fetch`.

Tests show a large speed-up is possible when processing a large number of inputs (1718 in this case):

**Before**
```
xbps-query -Rp pkgver -s python  0.18s user 0.02s system 99% cpu 0.199 total
cut -d: -f1  0.00s user 0.00s system 0% cpu 0.198 total
xargs -n1 xbps-uhelper getpkgname > /dev/null  0.49s user 1.17s system 94% cpu 1.761 total
```
**After**
```
xbps-query -Rp pkgver -s python  0.17s user 0.03s system 99% cpu 0.200 total
cut -d: -f1  0.00s user 0.00s system 0% cpu 0.200 total
xargs ./bin/xbps-uhelper/xbps-uhelper getpkgname > /dev/null  0.00s user 0.00s system 1% cpu 0.197 total
```
(note the lack of `-n1` in the second one)

- add getname and getversion actions

These actions are kind of "meta" actions, combining getpkgdepname and getpkgname (and the respective version actions) so that a list of mixed pkgvers and package patterns can be interpreted. This uses the internal format check of `xbps_pkgpattern_{name,version}()` to allow for a fallback to `xbps_pkg_{name,version}()` for exact versions, then falls back to displaying an error message if that also fails.

----

this will also be somewhat intertwined with an `xbps-uhelper` manpage, like in #533

### DIFF
--- a/bin/xbps-uhelper/main.c
+++ b/bin/xbps-uhelper/main.c
@@ -49,17 +49,17 @@ usage(void)
 	"    real-version, arch, getsystemdir\n"
 	"\n"
 	"  Action arguments:\n"
-	"    binpkgarch\t<binpkg>\n"
-	"    binpkgver\t<binpkg>\n"
+	"    binpkgarch\t<binpkg> ...\n"
+	"    binpkgver\t<binpkg> ...\n"
 	"    cmpver\t\t<instver> <reqver>\n"
-	"    getpkgdepname\t<string>\n"
-	"    getpkgdepversion\t<string>\n"
-	"    getpkgname\t\t<string>\n"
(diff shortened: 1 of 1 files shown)


leahneukirchen:
Useful!

classabbyamp:
> Useful!

would this be helpful for any xtools? I see `xpkg` uses sed at the moment

Issue #445, opened 2021-12-26
Use correct SI units in output of `xbps-install` and `xbps-remove`


camoz:
- Package size numbers (and maybe other sizes) are calculated/displayed in units based on powers of 2 (e.g. `KiB`), but the wrong unit is displayed, namely units based on powers of 10 (e.g. `KB`).
- There should be a space between the number and the unit suffix which is 1) standard according to SI and 2) more readable.

PR #542, opened 2023-02-20
format strings for library and xbps-query

Duncaen:
I'm not final on the api, don't really like the function names but couldn't come up with anything better.

Format strings are inspired by pythons `f""` strings:

```
$ xbps-query -F '{pkgname:20}\n' -m | head
               Carla
         ImageMagick
        LuaJIT-devel
             MEGAcmd
              PolyMC
          SDL2-32bit
          SDL2-devel
            SDL2_gfx
      SDL2_gfx-32bit
          SDL2_image
$ xbps-query -F '{pkgver:-30} {installed_size:10h}\n' -m | head
Carla-2.5.1_2                        79MB
ImageMagick-7.1.0.62_1              398KB
LuaJIT-devel-2.1.0beta3_2           898KB
MEGAcmd-1.4.0_1                    1300KB
PolyMC-6.3_1                           0B
SDL2-32bit-2.26.3_1                1841KB
SDL2-devel-2.26.3_1                6150KB
SDL2_gfx-1.0.4_2                     72KB
SDL2_gfx-32bit-1.0.4_2               83KB
SDL2_image-2.6.2_1                  127KB
$ xbps-query -F '0x{installed_size:8zX}\n' -m | head
0x04F62C86
0x000636AA
0x000E08BA
0x0014512A
0x00000000
0x001CC564
0x006019F2
0x00012075
0x00014D9C
0x0001FA1D
```


lib/format.c
```diff
@@ -0,0 +1,470 @@
...
+	fprintf(fp, "%s", str);
+	if (spec->align == '<' && spec->width > (unsigned int)len) {
+		for (unsigned i = 0; i < spec->width - len; i++)
+			fputc(spec->fill, fp);
+	}
+	return 0;
+}
+
+int
+xbps_fmt_number(const struct xbps_fmt_spec *spec, int64_t d, FILE *fp)
```
classabbyamp, line 416: would octal also make sense here? might be nice if you want to print numerical file mode

classabbyamp: maybe there's also some kind of special case that could be added later for doing strmode() on the value? (thinking about how i could use this for `query --long -f`)

lib/format.c
```diff
@@ -0,0 +1,470 @@
...
+			s.align = '>';
+			s.width -= 1;
+			fputc(buf[0], fp);
+			return xbps_fmt_string(&s, buf+1, len-1, fp);
+		}
+		break;
+	case 'x': len = snprintf(buf, sizeof(buf), "%" PRIx64, d); break;
+	case 'X': len = snprintf(buf, sizeof(buf), "%" PRIX64, d); break;
+	case 'h':
+		len = humanize_number(buf, spec->prec, d, "B",
```
classabbyamp, line 440: should #445 be considered here?

classabbyamp (commented):
interested in this for #541

lib/format.c
Duncaen: Makes sense, switched `humanize_numbers(3)` to the freebsd version that supports it. We can switch the other places that use `xbps_humanize_number` later.

lib/format.c
Duncaen: Switched the format to `{var!humanize[ ][.][width][minscale[maxscale]][i][:spec]}`, this leaves space for `{var!strmode[:spec]}` and allows to configure the conversions independent from the later string padding using the normal spec suffix for variables.
The default for just `!humanize` matches #445. 
```
$ xbps-query -F '{installed_size!humanize:8}\n' -m | head
  79 MiB
 398 KiB
 898 KiB
 1.3 MiB
 0.0 KiB
 1.8 MiB
 6.0 MiB
  72 KiB
  83 KiB
 127 KiB
```

Duncaen:
```
     -F, --format
         Format for list output.

         <substitution> ::= "{" variable ["!" conversion] [":" format] "}"
         <variable>     ::= [a-zA-Z90-9_-]

         <conversion>   ::= humanize | strmode

         -- Convert inode status information into a symbolic string
         <strmode> ::= "strmode"

         -- Format a number into a human readable form, the default is:`humanize .8Ki`:
         <humanize>     ::= "humanize" [space] [decimal] [width] [scale] [i]
           <space>      ::= " "          -- Put a space between number and the suffix.
           <decimal>    ::= "."          -- If the final result is less than 10, display
                                            it using one digit.
           <width>      ::= [0-9]+       -- Width of the output.
           <scale>      ::= multiplier   -- Minimum scale multiplier and optionally
                            [multiplier] -- Maxium scale multiplier.
           <multiplier> ::= "B"
                          | "K" -- kilo
                          | "M" -- mega
                          | "G" -- giga
                          | "T" -- tera
                          | "P" -- peta
                          | "E" -- exa
           <i>          ::= "i" -- Divide number with 1000 instead of 1024.

         <format>      ::= [[fill] align] [sign] [width] ["." precision] [type]
           <fill>      ::= <any char> -- The character to use when aligning the output.
           <align>     ::= "<"        -- Left align.
                         | ">"        -- Right align.
                         | "="        -- Left align with zero paddings after the sign.
           <sign>      ::= "+"        -- Add sign to positive and negative numbers.
                         | "-"        -- Add sign to negative numbers.
           <width>     ::= [0-9]+     -- The alignment width.
           <precision> ::= [0-9]+     -- Percision for numbers.
           <type>      ::= "d"        -- Decimal number.
                         | "o"        -- Octal number.
                         | "u"        -- Unsigned number.
                         | "x"        -- Hexadecimal with lowercase letters.
                         | "X"        -- Hexadecimal with uppercase letters.
```

lib/format.c
classabbyamp: nice! maybe short versions of the conversions could also be done? like `!h` for `!humanize` and `!m` for `!strmode`

bin/xbps-query/xbps-query.1
```diff
@@ -88,6 +88,49 @@ If the first character is not '/' then it's a relative path of
 .Ar rootdir .
 .It Fl d, Fl -debug
 Enables extra debugging shown to stderr.
+.It Fl F, Fl -format Ar format
+Format for list output.
+.Bd -literal
```
classabbyamp, line 93: nit: this is a giant block, maybe it make more sense to do
```roff
Format for list output. See
.Sx FORMAT STRINGS
for syntax.
```
and make it a section of the page

Duncaen:
```
FORMAT STRINGS
     Variables are package properties if not otherwise documented.  See
     PROPERTIES section for a list of available properties.

     As example a format string like:

           {pkgname:<30} {installed_size!humanize :>10}\n

     Would produce a list formatted like:

           libxbps                            304 KB
           xbps                               484 KB

     Format strings are parsed by the following EBNF:

     <grammar>      ::= (text | escape | substitution)*
     <text>         ::= [^\{}]+        -- literal text chunk
     <escape>       ::= "\" [abfnrtv0] -- POSIX-like espace sequence
                      | "\{" | "\}"    -- escaped "{" and "}"

     <substitution> ::= "{" variable ["!" conversion] [":" format] "}"
     <variable>     ::= [a-zA-Z0-9_-]

     <conversion>   ::= humanize | strmode

     -- Convert inode status information into a symbolic string
     <strmode> ::= "strmode"

     -- Format a number into a human readable form, the default is:`humanize .8Ki`:
     <humanize>     ::= "humanize" [space] [decimal] [width] [scale] [i]
       <space>      ::= " "          -- Put a space between number and the suffix.
       <decimal>    ::= "."          -- If the final result is less than 10, display
                                        it using one digit.
       <width>      ::= [0-9]+       -- Width of the output.
       <scale>      ::= multiplier   -- Minimum scale multiplier and optionally
                        [multiplier] -- Maxium scale multiplier.
       <multiplier> ::= "B"
                      | "K" -- kilo
                      | "M" -- mega
                      | "G" -- giga
                      | "T" -- tera
                      | "P" -- peta
                      | "E" -- exa
       <i>          ::= "i" -- Divide number with 1000 instead of 1024.

     <format>      ::= [[fill] align] [sign] [width] ["." precision] [type]
       <fill>      ::= <any char> -- The character to use when aligning the output.
       <align>     ::= "<"        -- Left align.
                     | ">"        -- Right align.
                     | "="        -- Left align with zero paddings after the sign.
       <sign>      ::= "+"        -- Add sign to positive and negative numbers.
                     | "-"        -- Add sign to negative numbers.
       <width>     ::= [0-9]+     -- The alignment width.
       <precision> ::= [0-9]+     -- Percision for numbers.
       <type>      ::= "d"        -- Decimal number.
                     | "o"        -- Octal number.
                     | "u"        -- Unsigned number.
                     | "x"        -- Hexadecimal with lowercase letters.
                     | "X"        -- Hexadecimal with uppercase letters.
```

classabbyamp:
> <i>          ::= "i" -- Divide number with 1000 instead of 1024.

this seems backwards to me? MiB would be the powers of 2 Mebibytes and MB would be the 10s based megabytes

Duncaen:
> > `<i>          ::= "i" -- Divide number with 1000 instead of 1024.`
> 
> this seems backwards to me? MiB would be the powers of 2 Mebibytes and MB would be the 10s based megabytes

Yes that was wrong, changed to:
> `       <i>          ::= "i" -- Use IEE/IEC (and now also SI) power of two prefixes.`

bin/xbps-query/xbps-query.1
```diff
@@ -277,6 +282,70 @@ expression wins.
...
+  <scale>      ::= multiplier   -- Minimum scale multiplier and optionally
+                   [multiplier] -- Maxium scale multiplier.
+  <multiplier> ::= "B" -- byte
+                 | "K" -- kilo
+                 | "M" -- mega
+                 | "G" -- giga
+                 | "T" -- tera
+                 | "P" -- peta
+                 | "E" -- exa
+  <i>          ::= "i" -- Use IEE/IEC (and now also SI) power of two prefixes.
```
classabbyamp, line 332:
```suggestion
  <i>          ::= "i" -- Use IEEE/IEC (and now also SI) power of two prefixes.
```

bin/xbps-query/xbps-query.1
```diff
@@ -277,6 +282,70 @@ expression wins.
...
 .El
+.Sh FORMAT STRINGS
+Variables are package properties if not otherwise documented.
+See
+.Sx PROPERTIES
+section for a list of available properties.
+.Pp
+As example a format string like:
+.Bd -offset indent -literal
+{pkgname:<30} {installed_size!humanize :>10}\\n
```
classabbyamp, line 293: is `pkgname` a property? if not, would it be possible to have it (and maybe `version`?) as a "pre-filtered" pkgver?

classabbyamp (commented):
lgtm

bin/xbps-query/xbps-query.1
Duncaen: `pkgname` is added by xbps "dynamically" at runtime, which kinda sucks but we could probably do that for version too.

bin/xbps-query/xbps-query.1
Duncaen: Copied that from https://man.freebsd.org/cgi/man.cgi?query=humanize_number&apropos=0&sektion=3&manpath=FreeBSD+14.0-CURRENT&arch=default&format=html which uses `IEE/IEC`, but `IEEE` makes more sense, no idea if there is a reason.

bin/xbps-query/xbps-query.1
classabbyamp: ah i see pkgname in the manpage now

Duncaen: maybe have `{pkgver!pkgname}` and `{pkgver!version}` or something like that, nut sure if this really matters, maybe I should just change this in the documentation to `pkgver`.

bin/xbps-query/xbps-query.1
classabbyamp: yeah I'm assuming they meant https://ieeexplore.ieee.org/document/9714443

bin/xbps-query/xbps-query.1
classabbyamp: well pkgname is listed in PROPERTIES, so it should be fine. adding version as a dynamic property is fairly minor imo, so could be done later.

Duncaen: Main issue with the `pkgname` prop is that for the pkgdb it is added when the pkgdb is loaded, for packages out of repodata it sets it when you use `xbps_repo_get_pkg`, so this may not be set for certain cases where it iterates of the packages in the repo.

I think xtraeme added this as a hack because the code converted pkgvers to pkgnames all over the place.

classabbyamp:
feature request for this: default values. would be useful for e.g. printing the user/group. maybe the syntax could be `{variable?default!conv:spec}`, and would be used if the value is `NULL`

bin/xbps-query/xbps-query.1
```diff
@@ -277,6 +282,73 @@ expression wins.
...
+<strmode> ::= "strmode"
+
+-- Format a number into a human readable form, the default is:`humanize .8Ki`:
+<humanize>     ::= "humanize" [space] [decimal] [width] [scale] [i]
+  <space>      ::= " "          -- Put a space between number and the suffix.
+  <decimal>    ::= "."          -- If the final result is less than 10, display
+                                   it using one digit.
+  <width>      ::= [0-9]+       -- Width of the output.
+  <scale>      ::= multiplier   -- Minimum scale multiplier and optionally
+                   [multiplier] -- Maxium scale multiplier.
```
classabbyamp, line 327:
```suggestion
                   [multiplier] -- Maximum scale multiplier.
```

bin/xbps-query/xbps-query.1
```diff
@@ -277,6 +282,73 @@ expression wins.
...
+
+<format>      ::= [[fill] align] [sign] [width] ["." precision] [type]
+  <fill>      ::= <any char> -- The character to use when aligning the output.
+  <align>     ::= "<"        -- Left align.
+                | ">"        -- Right align.
+                | "="        -- Left align with zero paddings after the sign.
+  <sign>      ::= "+"        -- Add sign to positive and negative numbers.
+                | "-"        -- Add sign to negative numbers.
+  <width>     ::= [0-9]+     -- The alignment width.
+  <precision> ::= [0-9]+     -- Percision for numbers.
```
classabbyamp, line 345:
```suggestion
  <precision> ::= [0-9]+     -- Precision for numbers.
```

bin/xbps-query/xbps-query.1
classabbyamp: fixed it and a couple others in https://github.com/void-linux/xbps/pull/541/commits/bc43563ad7bde173662a7998f5159215b9bab114

Chocimier:
Some form of replacement on values is needed to produce machine-readable output.

    $ xbps-query -F '{pkgname},"{short_desc}"\n' -C /etc/xbps.d -m | grep dino # not a csv
    dino,"Modern XMPP ("Jabber") Chat Client using GTK+/Vala"
    $ xbps-query -F '\{"{pkgname}":"{short_desc}"\}\n' -C /etc/xbps.d -m | grep dino # not a json
    {"dino":"Modern XMPP ("Jabber") Chat Client using GTK+/Vala"}

Maybe separate argument, like `--transalate '&&amp;'`,  `--translate \"\\\" --translate \\\\\\`

classabbyamp (approved):
approval for concept and (user-facing) functionality

classabbyamp:
```
$ xbps-query -F '\{{pkgname!json}:{run_depends!json}\}\n' -m
...
{"android-udev-rules":}
...
```
an incantation like this can create invalid json (hacky I know). I think all that needs adding is a default for arrays?

PR #549, opened 2023-03-29
xbps-uhelper: add verbose output for cmpver/pkgmatch

classabbyamp:
I can never remember which retval means which thing. This only prints if `-v`/`--verbose` is specified.

Examples:
```
$ ./bin/xbps-uhelper/xbps-uhelper -v cmpver 1 2 ; echo ret: $?
1 < 2
ret: 255
$ ./bin/xbps-uhelper/xbps-uhelper -v cmpver 1 1 ; echo ret: $?
1 = 1
ret: 0
$ ./bin/xbps-uhelper/xbps-uhelper -v cmpver 2 1 ; echo ret: $?
2 > 1
ret: 1
$ ./bin/xbps-uhelper/xbps-uhelper -v pkgmatch 'foo-1.0_1' 'foo>=0' ; echo ret: $?
foo-1.0_1 matches foo>=0
ret: 1
$ ./bin/xbps-uhelper/xbps-uhelper -v pkgmatch 'foo-1.0_1' 'foo<0.1_1' ; echo ret: $?
foo-1.0_1 does not match foo<0.1_1
ret: 0
```

it also seems that getting an error from pkgmatch is currently impossible


leahneukirchen:
useful!

Chocimier (approved):
Nice, can you add completion too?

    diff --git a/data/_xbps b/data/_xbps
    index 31587571..6b5e3153 100644
    --- a/data/_xbps
    +++ b/data/_xbps
    @@ -232,6 +232,7 @@ _xbps_uhelper() {
                    {-C,--config=-}'[Full path to configuration file]:config file:_files' \
                    {-d,--debug}'[Debug mode shown to stderr]' \
                    {-r,--rootdir=-}'[Full path to rootdir]:root dir:_files -/' \
    +               {-v,--verbose}'[Verbose messages]' \
                    {-V,--version}'[Show XBPS version]' \
                    '1:action:->actions' \
                    '*:: :->args' && ret=0

classabbyamp:
> Nice, can you add completion too?

ope forgot that, done (in a better way i think)

Duncaen:
What about printing this to stderr instead of stdout?

classabbyamp:
yes good idea

PR #571, opened 2023-09-17
data/xbps.conf: remove old mirrors

classabbyamp:


classabbyamp:
the remaining instance of `a-hel-fi.m.` will be replaced in #574

classabbyamp:
just going to do it all in #574

PR #582, opened 2024-01-10
Replace `xbps_file_hash_check_dictionary`

Duncaen:
Updated version of https://github.com/void-linux/xbps/pull/359, same results slightly different with less changes and ready to merge.

`xbps_file_hash_check_dictionary` was called for every file that is
getting unpacked, because the files list is an array it iterated over
the whole files array to find the matching file.
With a lot of files this is really slow and a lot of time was spend
in locking the proplib array and iterating over it.

Time from my Ryzen 3700X system with nvme disk updating
texlive-fontsextra with files 86375:

```
time xbps-install -y texlive-fontsextra
    6m34.61s real     6m27.71s user     0m03.95s system
```

And with this patch:

```
time xbps-install -y texlive-fontsextra
    0m08.40s real     0m07.34s user     0m00.98s system
```


classabbyamp:
on an RPi1B+:

before:
```
time xbps-install -y rpi-kernel rpi-kernel-headers
    34.13s user 78.40s system 53% cpu 3:29.33 total
```
after:
```
time xbps-install -y rpi-kernel rpi-kernel-headers
    33.75s user 79.38s system 53% cpu 3:30.23 total
```
i guess that package doesn't do much iterating

PR #587, opened 2024-02-10
data/_xbps_src: update completions

classabbyamp:
many flags were missing, also updates for new flags:
- host arch (void-linux/void-packages#46263)
- verbose (void-linux/void-packages#48637)


leahneukirchen:
Yes, noticed that too. Thanks!

Issue #592, opened 2024-03-18
xbps-install -u xbps may leave system in an unusable state


oreo639:
When updating through xbps, if the repository has an update for xbps available, then the only install/update action that may be performed is `xbps-install -u xbps`.

Since this may sometimes result in a situation where xbps may be unable to be updated since one of its dependencies may break their dependents, the decision was made to have `xbps-install -u xbps` ignore all missing dependency and missing shlibs errors:
https://github.com/void-linux/xbps/blob/master/lib/transaction_ops.c#L276-L285

This is problematic since with glibc 2.38, libcrypt.so.1 was dropped. In order to upgrade xbps, glibc must be updated, however, the glibc upgrade may break glibc dependents, which gets ignored.

The comment in the transaction_ops.c file suggests that you can just make sure to do `xbps-install -Su` afterwards and the problem will be resolved, however, if something like PAM gets broken, then now you now are going to have a real hard time logging in or changing user (`sudo`, for example, won't work anymore).

My question is, is this really the best way to handle this or is there a more intelligent way to handle xbps self upgrades.
(I know it is too late to fix this for the old ISOs but if we can fix this for the future, then that would be nice)

rubyFeedback:
Yes, that is also confusing. In my case binutils-doc could not be installed, so the rest of it was not installed either (I wanted to install gcc; and in my case the problem was that my temp dir did not exist). I think the xbps installer makes too many wrong assumptions by default. Some components of the linux stack require more care though as they can render the system unusable.

Issue #580, opened 2024-01-04
shlib errors on partial upgrade bypassed by multilib


oreo639:
When performing shlib checks before upgrading, libraries provided by 32bit packages are treated the same as libraries provided by 64bit packages. This can result in system breakage on partial upgrades without warning as noticed on the r/voidlinux subreddit: https://www.reddit.com/r/voidlinux/comments/18r0swm/psa_libxcrypt_updates/

I verified this behavior using the following:

1. Download [testpkgs.zip](https://github.com/void-linux/xbps/files/13830298/testpkgs.zip)
2. Run `./setup.sh`
3. In the cloned void-packages repo run:
```
./xbps-src binary-bootstrap
./xbps-src -m masterdir-i686 binary-bootstrap i686
./xbps-src pkg libhelloworld
./xbps-src -m masterdir-i686 pkg libhelloworld
./xbps-src pkg test1
xi test1
```
4. You can run `test1` and it will output `hello world 7` as expected.
5. After that, modify the libhelloworld package to not install libhelloworld.so.1 and build it as `version=2` (do not build the 32bit package with this change).
6. When doing `xi -u`, you will receive an shlibs error as expected.
7. Run `xi libhelloworld-32bit`
8. Now run `xi -u` again, you will not receive no warnings and the upgrade will proceed as normal.
9. Confirm the breakage of `test1` as running it now outputs `test1: error while loading shared libraries: libhelloworld.so.1: cannot open shared object file: No such file or directory`

PR #597, opened 2024-04-15
[RFC] lib: don't allow xbps self-update to bypass integrity checks

oreo639:
cc: @Duncaen

Right now, this is just a dumb implementation of the fix to get feedback, this does not remove the `xbps_autoupdate()` mechanism.
I'm concerned that leaving the autoupdate mechanism as-is may result in a situation where neither updating xbps or doing a full system upgrade is possible, would that block, for example, the builders (since they can't install anything).
However I am not sure what to replace it with, if anything? Ofc if you prefer to go a different route (e.g. static linking) please let me know.

You can test this using libhelloworld and test1 from https://github.com/void-linux/xbps/issues/580 and then having xbps depends on `libhelloworld>=<helloworld_version>` where helloworld_version would be 1 for the first build and 2 for the second ofc.

Closes: https://github.com/void-linux/xbps/issues/592


oreo639:
I made an alternative version where the self-update check is kept, but just made into a warning if that is preferable?
https://github.com/oreo639/xbps/commit/8cb137a49c1a6eff83eb960677955c87cf18c368

Duncaen:
Seems reasonable to make it a warning.

PR #595, opened 2024-03-27
CI: update CodeQL workflow

Duncaen:

PR #598, opened 2024-05-23
include: Do not generate xbps.h.in

meator:
xbps.h.in makes no mention of `@@VERSION@@`, so xbps.h.in gets copied verbatim to xbps.h. This adds unnecessary complexity and indirection.


include/Makefile
```diff
@@ -4,7 +4,7 @@ INCS =	xbps.h
 
 .PHONY: all
 all:
-	sed -e "s|@@VERSION@@|${VERSION}|g" ${INCS}.in > ${INCS}
+	@:
```
meator, line 7: I'm not sure if this is the right solution. Removal of the `all` target would lead to `install` becoming the default one, which is not desirable. I have replaced sed with the "do nothing" `@:` statement instead.

meator:
Check error is caused by faulty checks: #595

PR #605, opened 2024-12-03
bin/xbps-create: ignore provides file

classabbyamp:
will soon be used by xbps-src for collecting provides entries


classabbyamp:
no longer needed with https://github.com/void-linux/void-packages/pull/53401 and a refactor in https://github.com/void-linux/void-packages/pull/53326

PR #617, opened 2025-01-02
.github/workflows/codeql: update actions, containers, fix error

classabbyamp:
bash is needed for some scripts codeql runs


classabbyamp:
merged in #616

PR #625, opened 2025-05-19
lib: fix virtual package lookup performance

Duncaen:
Instead of having a dict with pkgvers as keys and iterating over all keys to match virtual packages, change the map to have pkgnames as keys, which hold another dict with all the old keys and values for a given package name.

This significantly speeds up XBPS, since for every package it would iterate over the whole virtual package map. With recent changes to xbps-src, where it generates cmd: and py3: virtual packages, the number of virtual packages in the map has been growing a lot.


classabbyamp (approved):
seems sound to me

PR #658, opened 2025-11-25
configure: automatically detect pkg-config to use, like CC

averyterrel:
Currently, when cross compiling, available libraries are incorrectly detected. This fixes that, and allows changing pkg-config implementations.

### DIFF
--- a/configure
+++ b/configure
@@ -189,6 +189,28 @@ else
 	echo "Using compiler $CC"
 fi
 
+# pkg-config is required to know dependencies for static linking.
+if [ -z "$PKG_CONFIG" ]; then
+	printf "Looking for pkg-config... "
+	for b in $TARGET- ""; do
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -189,6 +189,27 @@ else
...
+			if type $b$pkg_config >/dev/null 2>&1; then
+				PKG_CONFIG=$b$pkg_config
+				echo "$PKG_CONFIG"
+				break
+			fi
+		done
+		[ -n "$PKG_CONFIG" ] && break
+	done
+	if [ -z "$PKG_CONFIG" ]; then
+		echo "not found, exiting."
```
classabbyamp:
```suggestion
		echo
		echo "no suitable pkg-config tool found - aborting" >&2
```
to match the `$CC` check

PR #662, opened 2025-12-09
multi repo staging

Duncaen:


bin/xbps-rindex/index-add.c
```diff
@@ -281,86 +320,342 @@ index_add_pkg(struct xbps_handle *xhp, xbps_dictionary_t index, xbps_dictionary_
...
+		for (unsigned i = 0; i < nstates; i++) {
+			r = unstage(&states[i]);
+			if (r < 0)
+				goto err;
+		}
+	}
+
+	for (unsigned i = 0; i < nstates; i++) {
+		struct repo_state *state = &states[i];
+		// XXX: kinda useless to print for all repos, also print stage?.
```
classabbyamp: imo it would be useful to print the stage count too. maybe this could be prefixed by the repo path? 

```
/path/to/repo: index: %u packages registered.
```

bin/xbps-rindex/index-add.c
```diff
@@ -281,86 +320,342 @@ index_add_pkg(struct xbps_handle *xhp, xbps_dictionary_t index, xbps_dictionary_
...
+	return 0;
+}
+
+static int
+print_staged_packages(struct repo_state *states, unsigned nstates)
+{
+	xbps_object_iterator_t iter;
+	xbps_object_t keysym;
+
+	// XXX: should this really print all staged packages every time?
```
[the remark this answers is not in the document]
Duncaen: The current version prints everything since there is nothing really tracking what was added last, it just always adds to the stage then runs the staging check and unstages if possible. I could probably just add another `xbps_array_t` with all the newly added package versions and just print them as is.

bin/xbps-rindex/index-add.c
Duncaen: The comment about it being "useless" is that previously `xbps-rindex` could only add packages to one repo at a time and so printing the stats for just that repo that was being modified would've made sense. But theoretically with this change unless I'm missing something you could add all packages in different repositories with one xbps-rindex call.

So i guess it depends whether that is allowed or not.

Not sure if the distinction makes sense, since the xbps-rindex call that unstages the repo, unstages not just the repo we added to in that case printing stats for those repos too makes sense. Maybe its good to track and print stats only for modified repos.

classabbyamp: I think it would be ok (and more ergonomic and expected by the user) to allow adding packages to multiple repos in one invocation, especially when the repo can be modified during staging checks anyways

Duncaen:
```
Executing command [ xbps-rindex -v -R repo1 -R repo2 -a repo1/A-1.0_2.noarch.xbps ]
Fail: stdout not empty
--- /dev/null   2025-12-08 15:08:46.234592808 +0000
+++ /tmp/kyua.XduQkk/2/work/check.fkj0Ml/stdout 2025-12-09 20:37:00.678680807 +0000
@@ -0,0 +1,5 @@
+Inconsistent shlibs:
+  libruby.so.3.3 (provided by: ruby; used by: B)
+repo1: stage: added `A-1.0_2' (noarch)
+repo1: index: 5 packages registered.
+repo1: stage: 3 packages registered.
```

with the latest fixup, it will only print the newly added package and only print stats for repos that have been modified.

PR #670, opened 2026-01-07
bin/xbps-checkvers: unhardcode cache file version

ojab:
Right now cache file versioning doesn't follow xbps version, make it so.

Fixes: 34a1ab51d07996738529639114ace5a576b3fd62

### DIFF
--- a/bin/xbps-checkvers/main.c
+++ b/bin/xbps-checkvers/main.c
@@ -791,7 +791,7 @@ main(int argc, char **argv)
 		free(tmp);
 	}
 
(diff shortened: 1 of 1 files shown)


classabbyamp:
it was probably hard-coded because the format of the output hasn't changed since 0.58, just like how the pkgdb plist is versioned 0.37

ojab:
Yeah, there are no actual issues caused by hardcoded version and I'm mostly annoyed because of the `0.58` file in the FS without xbps-0.58 installed.
This change wouldn't break anything and it wouldn't be possible to forgot to do it if really needed.

Anyway, I'm fine with closing the PR and less annoyed since I know where `0.58` is coming from.

PR #685, opened 2026-04-16
data: add basic fish completions for xbps-src

balejk:
I have not been able to figure out yet how to make fish automatically look for completions for programs not in PATH so this currently needs to be sourced explicitly.


balejk:
Apparently this was disabled on purpose in https://github.com/fish-shell/fish-shell/commit/b2eea4b46f883b9a73b7d5a2cbd21b1d16e9736f. It could probably be worked around by sourcing this explicitly in a shipped file placed under /etc/fish/conf.d.

